PR #2, opened 2017-01-17
kubeadm-nspawn: Add a CLI for spawning the kubeadm dev environments

nhlfr:
A small CLI tool for downloading CentOS container image, spawning the containers, configuring a Docker daemon on them and setting up k8s cluster by kubeadm (via Ansible).

---

TODO:
- [x] Spawn "nodes" (systemd-nspawn containers)
- [x] Use CNI for providing networking to the nspawn containers
- [x] Bootstrap nodes - set up Docker and OpenSSH daemon
- [x] Distribute binaries / hyperkube image built from k8s sources on host
- [x] Run kubeadm on the nodes (machinectl shell, or machinectl login, or SSH, depends on what solution will work)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,15 @@
+all: install-vendor
+	go build -o cni-noop ./cmd/cni-noop
+	go build -o cnispawn ./cmd/cnispawn
+	go build -o kubeadm-systemd ./cmd/kubeadm-systemd
+
+check-glide-installation:
+	@which glide || go get -u github.com/Masterminds/glide
+
+install-vendor: check-glide-installation
+	glide install --strip-vendor
+
+update-vendor: check-glide-installation
+	glide update --strip-vendor
+
+.PHONY: check-glide-installation install-vendor update-vendor all
--- a/cmd/cnispawn/cnispawn.go
+++ b/cmd/cnispawn/cnispawn.go
@@ -1,0 +1,36 @@
+/*
+Copyright 2017 Kinvolk GmbH
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"flag"
+	"log"
+
+	"github.com/kinvolk/kubeadm-systemd/pkg/cnispawn"
+)
+
+var (
+	path       = flag.String("path", "", "Path to the container to run")
+	background = flag.Bool("background", true, "Spawn the container in background")
+)
+
+func main() {
+	flag.Parse()
+	if err := cnispawn.RunContainer(*path, *background); err != nil {
+		log.Fatal(err)
+	}
+}
--- a/cmd/kubeadm-systemd/kubeadm-systemd.go
+++ b/cmd/kubeadm-systemd/kubeadm-systemd.go
@@ -1,0 +1,165 @@
+/*
+Copyright 2017 Kinvolk GmbH
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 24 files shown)


Makefile
```diff
@@ -0,0 +1,5 @@
+export GO15VENDOREXPERIMENT=1
```
alban: do we still need to support Go 1.5 nowadays?

cmd/cnispawn/cnispawn.go
```diff
@@ -0,0 +1,36 @@
+/*
+Copyright 2016 Kinvolk GmbH
```
alban: 2017

cmd/kubeadm-systemd/kubeadm-systemd.go
```diff
@@ -0,0 +1,78 @@
+/*
+Copyright 2016 Kinvolk GmbH
```
alban: 2017

pkg/ansible/executor.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2016 Kinvolk GmbH
```
alban: 2017

pkg/ansible/executor.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	ansiblePath, err := exec.LookPath("ansible-playbook")
+	if err != nil {
+		return err
+	}
+
+	args := []string{ansiblePath, "-i", "inventory", "ansible/site.yml"}
+
+	pid, err := syscall.ForkExec(ansiblePath, args, &syscall.ProcAttr{
```
alban: Any reason not to use https://golang.org/pkg/os/exec/?

pkg/ansible/template.go
```diff
@@ -0,0 +1,53 @@
+/*
+Copyright 2016 Kinvolk GmbH
```
alban: 2017

pkg/bootstrap/container.go
```diff
@@ -0,0 +1,95 @@
...
+
+	if _, err := io.Copy(out, in); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func chmodScript(path string) error {
+	return os.Chmod(path, 0645)
```
alban: -rw-r--r-x ?

Only "other" can execute it?

pkg/bootstrap/ssh.go
```diff
@@ -0,0 +1,120 @@
...
+	if err := pem.Encode(privKeyFile, privKeyPem); err != nil {
+		return err
+	}
+
+	pubKey, err := ssh.NewPublicKey(&privKey.PublicKey)
+	if err != nil {
+		return err
+	}
+
+	if err := ioutil.WriteFile(pubKeyPath, ssh.MarshalAuthorizedKey(pubKey), 0655); err != nil {
```
alban: any reason not to use 0600 like for other keys in ~/.ssh/?

pkg/cnispawn/spawn.go
```diff
@@ -0,0 +1,64 @@
...
+	args := []string{
+		systemdNspawnPath,
+		"--capability=cap_audit_write,cap_audit_control,cap_sys_admin,cap_sys_chroot,cap_net_admin,cap_setfcap",
+		"--bind=/sys/fs/cgroup",
+		"-bD",
+		path,
+	}
+
+	if background {
+		_, err := syscall.ForkExec(systemdNspawnPath, args, &syscall.ProcAttr{
```
alban: wouldn't nspawn be killed if the terminal is closed?

pkg/bootstrap/ssh.go
nhlfr: Not really, but the same thing happens when you generate keys with `ssh-keygen` - only the private stuff gets 0600 mode, while the pub key permissions are more open.

But no problem I can change it, probably no difference for us.

alban: ah, I didn't know that the public keys got a more public permissions when using ssh-keygen. My public keys are 0600 too, maybe I did that myself, I don't remember. No strong opinion here.

pkg/cnispawn/spawn.go
nhlfr: It wouldn't. Not giving stdin to the nspawn process does the trick. And the nspawn process gets owned by systemd (from what I see in `pstree`).

nhlfr: However, it seems that it may be killed when exiting user session. So I will try to use `systemd run` here.

cmd/cnispawn/cnispawn.go
```diff
@@ -0,0 +1,36 @@
+/*
```
nhlfr: In kubernetes they use `/* */`, I wanted to be consistent with them.

nhlfr:
I'm going to split it in smaller PR-s

PR #3, opened 2017-01-17
Add README

### LINKED ISSUE #1: Self-hosted vs. Systemd
Hey @nhlfr! @blixtra pointed me here and I am wondering what the direction of this would be. We've done a lot of systemd/kubernetes in the past although we are slowly moving away from this. So I am happy to share some thoughts about this.

Obviously I don't know much about the scope of this tool so please ignore any false assumptions.

I'll try to list a few options to Integrate with systemd here:
1) generating the units to run a kubernetes on systemd. using either go binaries or rkt/runc to start container images with kubernetes binaries. using systemd-run to run those units right away.
2) using systemd-nspawn to start os images on a host that then use kubeadm to create a cluster
3) using systemd to start a kubelet and then start a self-hosted cluster rendered with something like bootkube. There is not much need to have something else than a kubelet if you go for the self-hosted kubernetes and that is what we do nowadays more or less.

You see I am just guessing and curious. Can you tell me more about your plans? And maybe we can help finding a good direction that is useful for the community.

Also pinging the kubeadm developers in case you haven't talked to them before @errordeveloper @lukemarsden @pires

nhlfr:
README which explains the goals, architecture and motivation behind the kubeadm-systemd project.

Fixes #1


nhlfr:
Merging, but if you have specific questions which I need to answer in README, please feel free to comment on the issue.

README.md
```diff
@@ -0,0 +1,60 @@
+# kubeadm-systemd
+
+## Introduction
+
+__kubeadm-systemd__ is a tool for creating a multi-node Kubernetes cluster
+on the single machine, created mostly for developers __of__ Kubernetes.
+
+It aims to be as similar to the solutions recommendend for the production
```
alban: typo on recommendend

README.md
```diff
@@ -0,0 +1,60 @@
...
+
+![Architecture Diagram](architecture.png?raw=true "Architecture")
+
+kubeadm-systemd uses the following third-party components to
+achieve its goal:
+
+### CNI
+
+Raw systemd-nspawn needs systemd-networkd for automanaging the networking
+for containers, which is bad for developers trying to use nspawn on their
```
alban: Is it true? systemd-nspawn has plenty of --network-* parameters.

README.md
```diff
@@ -0,0 +1,60 @@
...
+laptops - networkd doesn't provide necessary features for desktop systems
+where Network Manager is usually used. That's why we are using CNI as a
+tool for providing networking for nspawn.
+
+The integration between CNI and systemd-nspawn is made by our binary
+called __cnispawn__ which creates a network namespace, executes a CNI
+plugin on that, and then runs systemd-nspawn inside that namespace.
+By default, systemd-nspawn doesn't create its own network namepsaces,
+so the container is succesfully running inside the namespace we
```
alban: typo on succesfully

README.md
nhlfr: They don't work at all if you're using NetworkManager.

alban: because no dhcp server to assign IPs to the spawned containers?

README.md
blixtra: I'd add here that the focuses is currently on developers of k8s; leaving room for the focus to broaden later.

README.md
blixtra: networkd can do what you need but there is no real user interface for it. I'd say something like this...

We chose to use CNI to provide networking for systemd-nspawn as it's what is used in kubernetes itself and system-networkd, is currently trick to use.

nhlfr: @alban probably. Maybe there is a way to use some dhcp server to get `--network-*` options, but it sounds worse to me than just using CNI.

README.md
```diff
@@ -0,0 +1,60 @@
...
+  are using for setting up k8s clusters. That brings a very huge
+  risk that a developer _of_ k8s may be unable to reproduce some
+  bugs or issues which happen in clusters used in production.
+* [kubernetes-dind-cluster]{https://github.com/sttts/kubernetes-dind-cluster}
+  - it works great and does a great job in bringing multi-node clusters
+  for developers. But still, it uses its own way of creating the
+  cluster. And also, in our opinion, Docker isn't very good tool
+  for simulating the nodes, since it's an application container
+  engine, not an operaring system container engine (like
+  systemd-nspawn).
```
blixtra: How about...

But, as an app container engine, Docker does not simulate an operating system, thus providing a very different environment to what runs on production nodes.

README.md
```diff
@@ -0,0 +1,60 @@
...
+* [kubeadm-dind-cluster]{https://github.com/Mirantis/kubeadm-dind-cluster}
+  - it does a great job with using kubeadm, but still, it uses Docker
+  instead of any other container engine which is designed for
+  containerizing the whole OS, not an application. Also, we prefer
+  to maintain a code for doing such complicated things, instead of
+  huge shell scripts.
```
blixtra: it does a great job of making use of kubeadm. But as it uses Docker, it has the same issue as mentioned above; not simulating production nodes. Also, we prefer have the complexity of setting this up inside of Go code instead of shell scripts.

PR #4, opened 2017-01-17
README: Fix several typos and sentences

### LINKED ISSUE #1: Self-hosted vs. Systemd

nhlfr:
Fixes #1

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,9 @@
 
 __kubeadm-systemd__ is a tool for creating a multi-node Kubernetes cluster
 on the single machine, created mostly for developers __of__ Kubernetes.
+However, it may broaden its focus later.
 
-It aims to be as similar to the solutions recommendend for the production
+It aims to be as similar to the solutions recommended for the production
 clusters as possible.
 
 ## Architecture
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,17 +18,17 @@ achieve its goal:
 
 ### CNI
 
-Raw systemd-nspawn needs systemd-networkd for automanaging the networking
-for containers, which is bad for developers trying to use nspawn on their
-laptops - networkd doesn't provide necessary features for desktop systems
-where Network Manager is usually used. That's why we are using CNI as a
-tool for providing networking for nspawn.
+We chose to use CNI to provide networking for systemd-nspawn as it's what
+is used in kubernetes itself and system-networkd is currently tricky to use.
```
blixtra: system**d**-networkd

nhlfr: done

PR #22, opened 2017-05-18
Wrap docker-runc-current to avoid keyctl syscall

### LINKED ISSUE #21: Docker newer than v1.10.3 doesn't work
systemd-nspawn is blocking the `keyctl` syscall needed by runc

https://github.com/opencontainers/runc/issues/1369#issuecomment-286290644

robertgzr:
~~`cmd/runc-wrapper`~~ `cmd/nspawn-runc` wraps docker's `runc` as a custom runtime
and injects the `--no-new-keyring` flag when the create or run subcommands are executed.

Closes #21

### DIFF
--- a/cmd/nspawn-runc/nspawn-runc.go
+++ b/cmd/nspawn-runc/nspawn-runc.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2017 Kinvolk GmbH
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"log"
+	"os"
+	"os/exec"
+)
+
+var (
+	runcPath string = os.Getenv("NSPAWN_RUNC_BINARY_PATH")
+	logPath  string = os.Getenv("NSPAWN_RUNC_LOG_PATH")
+)
+
+func main() {
+	var newArgs []string
+
+	if logPath != "" {
+		fd, err := os.OpenFile(logPath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0660)
+		if err != nil {
+			log.Printf("error opening logs, skipping: %s", err)
+		}
+		log.SetOutput(fd)
+		defer fd.Close()
+	}
+
+	log.Printf("old args: %#v", os.Args[1:])
+
+	for _, a := range os.Args[1:] {
+		newArgs = append(newArgs, a)
+		if a == "create" || a == "run" {
+			newArgs = append(newArgs, "--no-new-keyring")
+		}
+	}
+
+	log.Printf("new args: %#v", newArgs)
+
+	if runcPath == "" {
+		var err error
+		runcPath, err = exec.LookPath("docker-runc")
+		if err != nil {
+			// unable to find default
+			log.Fatal(err)
+		}
+	}
+	cmd := exec.Command(runcPath, newArgs...)
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
+	cmd.Stdin = os.Stdin
+	if err := cmd.Run(); err != nil {
+		log.Fatal(err)
+	}
+}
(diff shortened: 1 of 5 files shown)


cmd/runc-wrapper/runc-wrapper.go
```diff
@@ -0,0 +1,45 @@
...
+		log.Println("error opening logs, skipping:", err)
+	}
+	log.SetOutput(fd)
+	defer fd.Close()
+
+	log.Printf("old args: %+v", os.Args[1:])
+
+	for _, a := range os.Args[1:] {
+		if a == "" {
+			a = "\"\""
```
alban: Is this necessary with the wrapper in Golang? I don't think that `exec.Command` removes empty parameters.

robertgzr: I tried it without that before, but Go also parses it as an empty argument and I need to explicitly tell it to write that into the new arguments...

First is before removing that block, second after:
```
2017/05/18 12:11:35 old args: [--console  --blabla sadlfkjldkfjlaksdfj]
2017/05/18 12:11:35 new args: [--console "" --blabla sadlfkjldkfjlaksdfj]
---
2017/05/18 14:54:07 old args: [--console  --blabla sadlfkjldkfjlaksdfj]
2017/05/18 14:54:07 new args: [--console  --blabla sadlfkjldkfjlaksdfj]
```

alban: Isn't this an artifact of the printf function, with the way it prints arrays?

robertgzr: You are right 😅 I just tested actually running it like this and it works.

robertgzr:
@alban Is there anything that needs to be done here or can we merge this?

cmd/runc-wrapper/runc-wrapper.go
```diff
@@ -0,0 +1,58 @@
...
+package main
+
+import (
+	"log"
+	"os"
+	"os/exec"
+)
+
+const (
+	runc        string = "/usr/libexec/docker/docker-runc-current"
```
alban: I think this path is Fedora-specific and it might be somewhere else on other distributions. (to check)

cmd/runc-wrapper/runc-wrapper.go
```diff
@@ -0,0 +1,58 @@
...
+	logLocation string = "/tmp/nspawn-runc-wrapper.log"
```
alban: This seems to be for debugging. Could we do that only if the user requests debugs?

Is it possible to have an environment variable that optionally sets up the log file? I don't know if the user can enable/disable an environment variable in `/etc/docker/daemon.json`...

robertgzr: yes it is possible to specify `"runtimeArgs"`

cmd/runc-wrapper/runc-wrapper.go
nhlfr: `/usr/libexec` is Fedora-specific, that's right.

robertgzr: `docker info --format '{{json .}}' | jq '.Runtimes.runc.path'` can tell us...

alban: It does not work for me:
```
$ sudo docker version
Client:
 Version:         1.12.6
 API version:     1.24
 Package version: docker-common-1.12.6-6.gitae7d637.fc25.x86_64
 Go version:      go1.7.4
 Git commit:      ae7d637/1.12.6
 Built:           Mon Jan 30 16:15:28 2017
 OS/Arch:         linux/amd64

Server:
 Version:         1.12.6
 API version:     1.24
 Package version: docker-common-1.12.6-6.gitae7d637.fc25.x86_64
 Go version:      go1.7.4
 Git commit:      ae7d637/1.12.6
 Built:           Mon Jan 30 16:15:28 2017
 OS/Arch:         linux/amd64

$ sudo docker info --format '{{json .}}'
flag provided but not defined: --format
See '/usr/bin/docker-current info --help'.
```

robertgzr:
* runc binary path can be set by env var
* by default we look for `docker-runc` in PATH
* made logging optional and location able to be specified with env var

cmd/nspawn-runc/nspawn-runc.go
```diff
@@ -0,0 +1,68 @@
...
+	logPath  string = os.Getenv("NSPAWN_RUNC_LOG_PATH")
+)
+
+func main() {
+	var newArgs []string
+
+	if logPath != "" {
+		fd, err := os.OpenFile(logPath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0660)
+		if err != nil {
+			log.Println("error opening logs, skipping:", err)
```
schu: Missing a `%v`

schu:
We agreed to merge this as is, @robertgzr will change/cleanup things in coming pull requests.

Issue #24, opened 2017-05-22
kubeadm-nspawn does not build because of CNI changes


alban:
```
$ make
...
go build -o kubeadm-nspawn ./cmd/kubeadm-nspawn
# github.com/kinvolk/kubeadm-nspawn/pkg/nspawntool
pkg/nspawntool/run.go:71: cniData.IP4 undefined (type types.Result has no field or method IP4)
```

IP4 was removed, see:
https://github.com/containernetworking/cni/pull/145/commits/befb95977ce2eb4a75aaff01f1356ef56beb271e#diff-34dfcb9f9c18e85cfeb1525b94b82572L81

alban:
I ran `make` following the instructions. This changed the version of cni used:
```diff
diff --git a/glide.lock b/glide.lock
index 84764ad..55c9585 100644
--- a/glide.lock
+++ b/glide.lock
@@ -1,27 +1,28 @@
-hash: 4fd9333c7f68ff1e551b0eb267aa915335782206a8501906e48ffc62e0400c1d
-updated: 2017-01-27T15:33:09.639904031+01:00
+hash: 72d0d4d2e5d4d7f27551716227fde0f5be2039524a16b66378da34a31d20730c
+updated: 2017-05-22T14:37:34.961254701+02:00
 imports:
 - name: github.com/containernetworking/cni
-  version: 5c3c17164270150467498a32c71436c7cd5501be
+  version: 8ea1605c26c50928e5fce74c2432d24c95ebd82c
   subpackages:
   - pkg/ns
   - pkg/types
.....
```

PR #25, opened 2017-05-22
Do not update vendor on every `make`

robertgzr:
Not sure if this solves #24

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,28 +1,27 @@
-all: vendor
+.PHONY: vendor all reset-cni
+.PHONY: clean clean-bins clean-rootfs clean-image clean-ssh-keys 
+
+all:
 	go build -o cni-noop ./cmd/cni-noop
 	go build -o cnispawn ./cmd/cnispawn
 	go build -o kubeadm-nspawn ./cmd/kubeadm-nspawn
 
+vendor: glide.lock | glide
+	glide --quiet install --strip-vendor
+glide.lock: | glide
+	glide update --strip-vendor
 glide:
 	@which glide || go get -u github.com/Masterminds/glide
 
-glide.lock: glide.yaml | glide
-	glide update --strip-vendor
-
-vendor: glide.lock | glide
-	glide install --strip-vendor
-
-.PHONY: glide vendor all
-
-clean:
-	rm -rf ./{kubeadm-nspawn,cni-noop,cnispawn}
-
+clean: clean-bins clean-rootfs clean-image clean-ssh-keys
+clean-ssh-keys:
+	rm -rf ./id_rsa*
+clean-bins:
+	rm -rf ./{cni-noop,cnispawn,kubeadm-nspawn}
 clean-rootfs:
 	sudo rm -rf kubeadm-nspawn-*
-
-clean-cni:
-	sudo rm -rf /var/lib/cni/networks/mynet/*
-	sudo iptables-restore < ~/iptables
-
 clean-image:
 	rm -rf rootfs.tar.xz
+
+reset-cni:
+	sudo rm -rf /var/lib/cni/networks/mynet


robertgzr:
@nhlfr should we add something like this: https://github.com/Mirantis/virtlet/blob/master/Makefile.am#L29-L32

Makefile
```diff
@@ -1,28 +1,25 @@
...
 	go build -o cni-noop ./cmd/cni-noop
 	go build -o cnispawn ./cmd/cnispawn
 	go build -o kubeadm-nspawn ./cmd/kubeadm-nspawn
 
-glide:
-	@which glide || go get -u github.com/Masterminds/glide
-
+vendor: glide.lock | glide
+	glide --quiet install --strip-vendor
 glide.lock: glide.yaml | glide
```
alban: I am not sure about this dependency on `glide.yml` from `glide.lock`. When git-cloning the kubeadm-nspawn repository for the first time, the vendor/ directory is not yet populated. The user has to run `glide install`. But running `make vendor` will not work because at the moment `glide.lock` is older than `glide.yaml` in the repo:

image

So this will trigger the rule `glide update` and it will likely not work for new users who just cloned the repository.

Maybe using a [order-only-prerequisites](https://www.gnu.org/software/make/manual/make.html#Prerequisite-Types) like this (see the position of the pipe):
```
glide.lock: | glide.yaml glide
```

Untested.

alban: I am unable to reproduce the error I had, so I am not sure if the explanation is correct...

@schu what do you think of changing the `|` as I suggested above?

Makefile
```diff
@@ -1,28 +1,25 @@
...
+clean-bins:
+	rm -rf ./{cni-noop,cnispawn,kubeadm-nspawn}
 clean-rootfs:
 	sudo rm -rf kubeadm-nspawn-*
-
 clean-cni:
-	sudo rm -rf /var/lib/cni/networks/mynet/*
-	sudo iptables-restore < ~/iptables
-
+	sudo rm -rf /var/lib/cni/networks/mynet
```
alban: In general, I don't think that `make clean` should delete files outside of the git repository.

Makefile
```diff
@@ -1,28 +1,25 @@
...
-	glide install --strip-vendor
-
-.PHONY: glide vendor all
-
-clean:
-	rm -rf ./{kubeadm-nspawn,cni-noop,cnispawn}
+.PHONY: vendor all clean
 
+clean: clean-bins clean-rootfs clean-cni clean-image clean-ssh-keys
+clean-ssh-keys:
```
alban: "clean-ssh-keys" is not a file, so it should be added in the `.PHONY` list.

See https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html#Phony-Targets:
> A phony target is one that is not really the name of a file

Makefile
robertgzr: I put that there because every time we are running `cnispawn` it creates a file this folder and increments the ip addresses of the spawned containers.

How are we going to clean this up instead?

Makefile
robertgzr: Why only `clean-ssh-keys`? Does the same not apply for the other clean targets?

alban: Yes, same apply for the others

Makefile
alban: I am fine to have a target for that, but I would prefer another name than "clean" because "clean" is a [standard target](https://www.gnu.org/prep/standards/html_node/Standard-Targets.html) that users could expect to have a standard meaning:

> clean: 
> Delete all files in the current directory that are normally created by building the program.

Any idea for another name?

robertgzr: `reset-cni` ?

Makefile
schu: Seems correct from a cursory read, didn't test.

Makefile
alban: Good name for me :)

robertgzr:
@alban @schu @nhlfr
Can this be merged?

Makefile
```diff
@@ -1,28 +1,27 @@
...
+.PHONY: clean clean-bins clean-rootfs clean-image clean-ssh-keys 
+
+all:
 	go build -o cni-noop ./cmd/cni-noop
 	go build -o cnispawn ./cmd/cnispawn
 	go build -o kubeadm-nspawn ./cmd/kubeadm-nspawn
 
+vendor: glide.lock | glide
+	glide --quiet install --strip-vendor
+glide.lock: | glide.yaml glide
```
schu: There's no targte `glide.yaml` defined.

robertgzr: Isn't that just making sure that the file exists?

schu: Sort of, yes. If

```
$ make vendor
make: *** No rule to make target 'glide.yaml', needed by 'glide.lock'.  Stop.
```

is what you want. The glide error is more descriptive though IMO:

```
[ERROR] Failed to find glide.yaml file in directory tree: Cannot resolve parent of /
```

i.e you could just remove `glide.yaml` here.

PR #31, opened 2017-06-13
Expose nspawn options in cnispawn

robertgzr:
This PR is in preparation for the move to container linux as default base image for the nodes.

### DIFF
--- a/pkg/cnispawn/spawn.go
+++ b/pkg/cnispawn/spawn.go
@@ -23,7 +23,7 @@ import (
 	"syscall"
 )
 
-func RunContainer(path string, background bool) error {
+func Spawn(background bool, nspawnArgs []string) error {
 	runtime.LockOSThread()
 
 	cniNetns, err := NewCniNetns()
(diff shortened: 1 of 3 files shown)


pkg/cnispawn/spawn.go
```diff
@@ -45,18 +45,18 @@ func RunContainer(path string, background bool) error {
...
 		"--bind=/sys/fs/cgroup",
 		"--bind-ro=/boot",
 		"--bind-ro=/lib/modules",
-		"-bD",
-		path,
+		"--boot",
+		"--notify-ready=yes",
 	}
 
+	args = append(args, nspawnArgs...)
```
robertgzr: @nhlfr what do you mean? Here I pass everything that you specify on the command line to nspawn

nhlfr: Ah, I missed that line, sorry.

robertgzr: The only problem I can see is that we take away the option to not have `--boot` for example... But I don't consider `cnispawn` to be useful on it's own anyway... at least not now

PR #41, opened 2017-06-15
Add cnibridge target to makefile

robertgzr:
This is the easiest possible solution for now.

Closes https://github.com/kinvolk/kubeadm-nspawn/issues/39

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,9 @@ all:
 	go build -o nspawn-runc ./cmd/nspawn-runc
 	go build -o kubeadm-nspawn ./cmd/kubeadm-nspawn
 
+cnibridge:
+	go get -u github.com/containernetworking/plugins/plugins/main/bridge
+
 vendor: glide.lock | glide
 	glide --quiet install --strip-vendor
 glide.lock: | glide


Makefile
```diff
@@ -7,6 +7,9 @@ all:
 	go build -o nspawn-runc ./cmd/nspawn-runc
 	go build -o kubeadm-nspawn ./cmd/kubeadm-nspawn
 
+cnibridge:
+	go get -u github.com/containernetworking/plugins/plugins/main/bridge
```
alban: This does not compile the vendored version but clone/update the CNI repository in `$GOPATH`.

robertgzr: As discussed with @nhlfr 
The problem is we strip the vendor folders from our dependencies https://github.com/kinvolk/kubeadm-nspawn/blob/master/Makefile#L15

I am not sure how to handle this... Maybe we could do a glide update/install without the `--strip-vendor` flag first, build the bridge plugin and then do run glide with the flag

alban: why do we have the flag `--strip-vendor` in the first place?

Also, `glide.lock` does not contain "github.com/containernetworking/plugins". It seems the vendoring and `glide.lock` need to be fixed as part of this PR.

robertgzr: it looks like it is necessary:
```
go build -o kubeadm-nspawn ./cmd/kubeadm-nspawn
# github.com/kinvolk/kubeadm-nspawn/pkg/distribution
pkg/distribution/registry.go:92: cannot use "github.com/kinvolk/kubeadm-nspawn/vendor/github.com/docker/go-connections/nat".PortMap literal (type "github.com/kinvolk/kubeadm-nspawn/vendor/github.com/docker/go-connections/nat".PortMap) as type "github.com/kinvolk/kubeadm-nspawn/vendor/github.com/docker/docker/vendor/github.com/docker/go-connections/nat".PortMap in field value
```

robertgzr: Maybe we can leave it for now and just do `go get -u ...`. We will still have to move our dependency on cni to their new repo structure.

After all it was not the code that changed but only the location of the bridge plugin...

robertgzr: I also don't think this should be in the Makefile... Maybe just adding this to the README is enough:
```
go get -u github.com/containernetworking/plugins/plugins/main/bridge
go get -u github.com/containernetworking/plugins/plugins/ipam/host-local
export CNI_PATH=$GOPATH/bin
```

robertgzr:
Closed here https://github.com/kinvolk/kubeadm-nspawn/issues/39

Issue #49, opened 2017-07-06
Init hangs due to a missing apiserver


dongsupark:
(This is a summary of offline discussion with @nhlfr )

"kubeadm-nspawn init" command is now able to connect to the correct api server. However, init still hangs with the following message:

<pre>
[discovery] Trying to connect to API Server "10.22.0.19:6443"
[discovery] Created cluster-info discovery client, requesting info from "https://10.22.0.19:6443"
[discovery] Failed to request cluster info, will try again: [the server has asked for the client to provide credentials (get configmaps cluster-info)]
...
[discovery] Failed to request cluster info, will try again: [Get https://10.22.0.19:6443/api/v1/namespaces/kube-public/configmaps/cluster-info: dial tcp 10.22.0.19:6443: getsockopt: connection refused]
...
# (hangs forever)
</pre>

I don't understand the error message "the server has asked for the client to provide credentials", as the token is already given correctly.

Inside the nspawn container for the master node:

<pre>
$ docker ps
CONTAINER ID        IMAGE                                                    COMMAND                  CREATED              STATUS              PORTS               NAMES
594868d709b1        weaveworks/weave-kube                                    "/home/weave/launc..."   48 seconds ago       Up 47 seconds                           k8s_weave_weave-net-r43wz_kube-system_a7eef4ba-6248-11e7-8a53-0a580a160013_0
bbf5d8ddeada        weaveworks/weave-npc                                     "/usr/bin/weave-npc"     48 seconds ago       Up 47 seconds                           k8s_weave-npc_weave-net-r43wz_kube-system_a7eef4ba-6248-11e7-8a53-0a580a160013_0
f1c6c00c1e54        gcr.io/google_containers/pause-amd64:3.0                 "/pause"                 48 seconds ago       Up 47 seconds                           k8s_POD_weave-net-r43wz_kube-system_a7eef4ba-6248-11e7-8a53-0a580a160013_0
55621f703ca5        gcr.io/google_containers/pause-amd64:3.0                 "/pause"                 50 seconds ago       Up 49 seconds                           k8s_POD_kube-proxy-1sf2v_kube-system_a6f01f65-6248-11e7-8a53-0a580a160013_0
[...]

PR #52, opened 2017-07-10
Update documentation

robertgzr:
Split up the README to only include general information and on how to run the
project.

Keep additional documentation in doc/

Needs #30 #45 #49 #54 #56  #57 

Closes #51

PR #53, opened 2017-07-11
Fix bugs in Vagrantfile and init scripts

dongsupark:
This PR contains multiple fixes to be able to run `kube-spawn` inside a Vagrant VM. The most critical one is the commit `"Vagrantfile: upgrade k8s to 1.7.0 for bugfixes"`.
 
* Vagrantfile: split out inline scripts into separate ones
* scripts: add kernel module for overlayfs
* scripts: add cni loopback plugin
* Vagrantfile: set vbguest config only if vbguest plugin is installed
* scripts: upgrade k8s to 1.7.0 for bugfixes
* kube-spawn: set the default k8s version to 1.7.0

This PR is basically a rebase of `"dongsu/devel-fix"` branch on top of the master branch. I'll update documents later when #52 got merged.

Fixes https://github.com/kinvolk/kube-spawn/issues/49

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -13,7 +13,9 @@ Vagrant.configure("2") do |config|
   config.vm.synced_folder ".", "/vagrant", disabled: true
   config.vm.synced_folder ".", "/home/vagrant/go/src/github.com/kinvolk/kube-spawn", create: true, type: "rsync"
 
-  config.vbguest.auto_update = false
+  if Vagrant.has_plugin?("vagrant-vbguest")
+    config.vbguest.auto_update = false
+  end
   config.vm.provider :virtualbox do |vb|
       vb.check_guest_additions = false
       vb.functional_vboxsf = false
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -21,40 +21,6 @@ Vagrant.configure("2") do |config|
       vb.customize ["modifyvm", :id, "--cpus", "1"]
   end
 
-  config.vm.provision "shell", env: {"GOPATH" => "/home/vagrant/go"}, privileged: false, inline: <<HERE
-echo 'Setting up correct env. variables'
-echo "export GOPATH=$GOPATH" >> /home/vagrant/.bash_profile
-echo "export PATH=$PATH:$GOPATH/bin:/usr/local/go/bin" >> /home/vagrant/.bash_profile
-
-echo 'Writing build.sh'
-source ~/.bash_profile
-if [[ ! -f /home/vagrant/build.sh ]]; then
-cat >>/home/vagrant/build.sh <<-EOF
-#!/bin/bash
-set -xe
-
-cd $GOPATH/src/github.com/kinvolk/kube-spawn
-
-go get -u github.com/containernetworking/plugins/plugins/main/bridge
-go get -u github.com/containernetworking/plugins/plugins/ipam/host-local
-
-make vendor all
-
-sudo machinectl show-image coreos || sudo machinectl pull-raw --verify=no https://alpha.release.core-os.net/amd64-usr/current/coreos_developer_container.bin.bz2 coreos
-
-sudo GOPATH=$GOPATH CNI_PATH=$GOPATH/bin ./kube-spawn --kubernetes-version=1.6.6 up --nodes 2 --image coreos
-sudo GOPATH=$GOPATH CNI_PATH=$GOPATH/bin ./kube-spawn --kubernetes-version=1.6.6 init
-EOF
-fi
-HERE
-
-  config.vm.provision "shell", inline: <<HERE
-echo 'Modifying environment'
-chown -R vagrant:vagrant /home/vagrant
-chmod +x /home/vagrant/build.sh
-setenforce 0
-systemctl stop firewalld
-sudo groupadd docker && sudo gpasswd -a ${USER} docker && sudo systemctl restart docker && newgrp docker
-usermod -aG docker vagrant
-HERE
+  config.vm.provision "shell", env: {"GOPATH" => "/home/vagrant/go"}, privileged: false, path: "scripts/vagrant-setup-env.sh"
+  config.vm.provision "shell", path: "scripts/vagrant-mod-env.sh"
 end
--- a/scripts/vagrant-mod-env.sh
+++ b/scripts/vagrant-mod-env.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -euo pipefail
+
+USER=vagrant
+HOME=/home/${USER}
+
+echo 'Modifying environment'
+chown -R ${USER}:${USER} ${HOME}
+chmod +x ${HOME}/build.sh
+setenforce 0
+systemctl stop firewalld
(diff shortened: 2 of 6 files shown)


cmd/kubeadm-nspawn/kubeadm-nspawn.go
```diff
@@ -162,7 +198,7 @@ func newKubeadmNspawnCommand() *cobra.Command {
 		},
 	}
 	cmd.Flags().BoolVarP(&printVersion, "version", "V", false, "output version information")
-	cmd.PersistentFlags().StringVarP(&k8srelease, "kubernetes-version", "k", "", "Kubernetes version to spawn")
+	cmd.PersistentFlags().StringVarP(&k8srelease, "kubernetes-version", "k", k8sStableVersion, "Kubernetes version to spawn")
```
robertgzr: Maybe we can have a special word to signal that we want to use local kubernetes. Something like `--kubernetes-version=dev` or setting the flag but passing an empty value.

dongsupark: Good point. Will handle the case of "dev" or "".

Vagrantfile
```diff
@@ -56,5 +59,7 @@ setenforce 0
 systemctl stop firewalld
 sudo groupadd docker && sudo gpasswd -a ${USER} docker && sudo systemctl restart docker && newgrp docker
 usermod -aG docker vagrant
+sudo modprobe overlay
+echo "131072" | sudo tee /sys/module/nf_conntrack/parameters/hashsize
```
dongsupark: Unfortunately it's necsssary. The hashsize patch is included in K8s 1.7. However kube-proxy doesn't run when hashsize=16k, but it starts to run when hashsize=32k or higher. That's probably because under Fedora 26 in the Vagrant VM, `net.nf_contrack_max` is set to 64k by default. So when the current hashsize isn't greater than (max / 4), say 16k, it doesn't escape from the code above, which results in an error due to netns restriction. That's what I think it's happening.

We could make a more brilliant guess here, but I just took a simple approach to increase the hashsize to 128k.

pkg/nspawntool/run.go
```diff
@@ -63,6 +63,7 @@ var defaultBinds = []string{
 	bindro + parseBind("$PWD/etc/kube_20-kubeadm-extra-args.conf:/etc/systemd/system/kubelet.service.d/20-kubeadm-extra-args.conf"),
 	// cni bins
 	bindro + path.Join(cnipath+":/opt/cni/bin"),
+	bindro + path.Join(gopath+"/bin/loopback:/opt/cni/bin/loopback"),
```
dongsupark: Oops, right. It's unnecessary. I'll remove the line 66.

Vagrantfile
robertgzr: Ah well seems alright... I would add that as a comment to the Vagrantfile it might be useful at one point to have that information somwhere.

schu:
Could you switch from inline to setup scripts included with `config.vm.provision "shell", path: "script.sh"`? Those could be shellcheck'ed then and are easier to read. And `set -euo pipefail`.

cmd/kubeadm-nspawn/kubeadm-nspawn.go
```diff
@@ -98,7 +129,16 @@ func newUpCommand() *cobra.Command {
...
 
 func runInit(cmd *cobra.Command, args []string) {
-	if k8srelease == "" {
+	if !checkK8sStableRelease(k8srelease) {
+		log.Printf("WARNING: K8s with version <%s is not compatible with kubeadm-nspawn.",
+			k8sStableVersion)
+		log.Printf("It's highly recommended to upgrade K8s to 1.7 or newer.")
+		// Print a kind warning, and continue to run.
+		// It should still allow kubeadm to run with Kubernetes <1.7,
+		// to be able to allow end users to flexibly handle various cases.
```
schu: I would factor this out into a function which can be used by both `runInit` and `runUp`

cmd/kubeadm-nspawn/kubeadm-nspawn.go
```diff
@@ -42,6 +44,26 @@ var (
...
+	c, err := semver.NewConstraint(">=" + k8sStableVersion)
+	if err != nil {
+		log.Printf("Cannot get constraint for >= %s: %v", k8sStableVersion, err)
+		return false
+	}
+
+	return c.Check(v)
+}
+
+func isK8sUpstream(k8srel string) bool {
```
robertgzr: I would call it `isDev()` or use `upstream` as the value to indicate that we mean non-release.

It might be only me but _upstream_ sounds like we're just building the official source. Maybe _local_ (and `isLocal()`) but I'm not sure...

Or a different approach. Could you reuse `checkK8sStableRelease` for that?

dongsupark:
@schu Done. Split out inline scripts into `scripts/vagrant-setup-env.sh` and `scripts/vagrant-mod-env.sh`, and applied patches on top of that. It will improve readability.

blixtra:
Needs to be rebased after merge of name change.

dongsupark:
@blixtra Rebased.
@robertgzr Renamed to `isDev()` as you suggested. Reusing other functions is not trivial, so I'll think about that in the future.

scripts/vagrant-mod-env.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -e
```
schu: `-e` not needed with `set -euo pipefail` below.

scripts/vagrant-setup-env.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash -e
+
+echo 'Setting up correct env. variables'
+echo "export GOPATH=$GOPATH" >> $HOME/.bash_profile
+echo "export PATH=$PATH:$GOPATH/bin:/usr/local/go/bin" >> $HOME/.bash_profile
+
+source ~/.bash_profile
+
+set -euo pipefail
```
schu: Can we move this up?

scripts/vagrant-setup-env.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash -e
+
+echo 'Setting up correct env. variables'
+echo "export GOPATH=$GOPATH" >> $HOME/.bash_profile
```
schu: shellcheck complains here (`SC2086`) and about a couple of other things.

scripts/vagrant-mod-env.sh
dongsupark: @schu Right. Removed `-e`.

scripts/vagrant-setup-env.sh
dongsupark: @schu Done. shellcheck doesn't complain any more.

scripts/vagrant-setup-env.sh
dongsupark: @schu `-e` and `-o pipefail` can be moved, but `-u` cannot, because we would otherwise meet an error `"PS1: unbound variable"` inside `.bash_profile`. So I let `-u` stay below `"source ~/.bash_profile"`, and added a comment.

PR #60, opened 2017-07-12
Vagrantfile, scripts: assign 2 CPUs and fix setup scripts

dongsupark:
In Vagrantfile, assign 2 CPUs for each VM to be able to build large source trees.
Also fix minor bugs in `vagrant-mod-env.sh`: Ignore return value of `setenforce`, as it always returns 1 if selinux is disabled. Also make sure that chmod runs only for a necessary directory k8s, instead of the whole home directory.

### DIFF
--- a/scripts/vagrant-mod-env.sh
+++ b/scripts/vagrant-mod-env.sh
@@ -6,9 +6,12 @@ USER=vagrant
 HOME=/home/${USER}
 
 echo 'Modifying environment'
-chown -R ${USER}:${USER} ${HOME}
+chown -R ${USER}:${USER} ${HOME}/go/src/github.com/kinvolk/kube-spawn/k8s
 chmod +x ${HOME}/build.sh
-setenforce 0
+
+# setenforce always returns 1 when selinux is disabled.
+# we should ignore the error and continue.
+setenforce 0 || true
 systemctl stop firewalld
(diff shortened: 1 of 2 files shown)


scripts/vagrant-mod-env.sh
```diff
@@ -6,9 +6,12 @@ USER=vagrant
 HOME=/home/${USER}
 
 echo 'Modifying environment'
-chown -R ${USER}:${USER} ${HOME}
+chown -R ${USER}:${USER} ${HOME}/go/src/github.com/kinvolk/kube-spawn/k8s
 chmod +x ${HOME}/build.sh
-setenforce 0
+
+# setenforce always returns 1 when selinux is disabled.
+# we should ignore the error value to continue to rest commands.
```
schu: Nit: "... to continue with the remaining commands" or, simpler, just "Ignore the error and continue".

dongsupark: @schu Fixed to "Ignore the error and continue".

schu: @dongsupark looks like the change wasn't actually in the commit?

dongsupark: @schu Sorry. Updated it again.

schu: Thanks :+1:

PR #62, opened 2017-07-14
pkg/nspawntool: mount /opt/cni/bin as read-write

dongsupark:
`/opt/cni/bin` needs to be mounted read-write, as additional files might need to be written into that directory inside nspawn containers.

Debugged by @iaguis

### DIFF
--- a/pkg/nspawntool/run.go
+++ b/pkg/nspawntool/run.go
@@ -32,6 +32,7 @@ import (
 )
 
 const bindro string = "--bind-ro="
+const bindrw string = "--bind="
 
 var (
 	gopath  string = os.Getenv("GOPATH")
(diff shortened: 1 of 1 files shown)


pkg/nspawntool/run.go
```diff
@@ -52,17 +53,17 @@ var (
 var k8sbinds []string
 var defaultBinds = []string{
 	// kube-spawn bins
-	"--bind=" + parseBind("$PWD/scripts:/opt/kube-spawn"),
+	bindrw + parseBind("$PWD/scripts:/opt/kube-spawn"),
 	bindro + parseBind("$PWD/kube-spawn-runc:/opt/kube-spawn-runc"),
 	// shared tmpdir
-	"--bind=" + parseBind("$PWD/tmp:/tmp/kube-spawn"),
+	bindrw + parseBind("$PWD/tmp:/tmp/kube-spawn"),
```
iaguis: this is a different change and, how did this work without it being `rw`?

dongsupark: @iaguis `/tmp/kube-spawn` was already mounted read-write. I just replaced `"--bind="` with `bindrw` to avoid duplications.

iaguis: ok, misread :+1:

PR #68, opened 2017-07-17
kube-spawn-runc: don't set Std{err,out,in} if we don't log

iaguis:
This avoids

```
fork/exec /run/torcx/bin/docker-runc: bad file descriptor
```

We haven't fully investigated why this happens but we don't really need
input/output for runc.

### DIFF
--- a/cmd/kube-spawn-runc/kube-spawn-runc.go
+++ b/cmd/kube-spawn-runc/kube-spawn-runc.go
@@ -59,9 +59,6 @@ func main() {
 		}
 	}
 	cmd := exec.Command(runcPath, newArgs...)
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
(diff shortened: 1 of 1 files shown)


cmd/kube-spawn-runc/kube-spawn-runc.go
```diff
@@ -59,9 +59,11 @@ func main() {
 		}
 	}
 	cmd := exec.Command(runcPath, newArgs...)
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
-	cmd.Stdin = os.Stdin
+	if logPath != "" {
```
robertgzr: Do we even need to keep these lines at all?

iaguis: Probably not, we already logged at this point. Thx.

Issue #66, opened 2017-07-14
No more space in btrfs volume /var/lib/machines


dongsupark:
While testing kube-spawn on Ubuntu/Debian, I saw a strange case of btrfs volume `/var/lib/machines` became suddenly read-only mounted, which obviously caused nspawn containers to fail completely.

`machinectl pull-raw` downloads Container Linux image `coreos.raw` under `/var/lib/machines`, which belongs to a storage pool backed by `/var/lib/machines.raw`. Initially it allocates exactly the same size in the pool as the original image. That's fine until kubeadm starts to write files inside nspawn containers, where at some point there's no more space for btrfs metadata. Even when a simple `df` command shows that more space is available, e.g. 30% free, metadata space for btrfs might be already full, as shown with `sudo btrfs filesystem usage /var/lib/machines`.

It's possible to avoid this by recreating the storage pool with a larger size, and re-execute nspawn containers manually. Then it works fine. But that's obviously not a solution. We need to think about a general way of solving this issue.

dongsupark:
It's not a distro-specific issue, as it also happens on Fedora.
Related issue: https://github.com/systemd/systemd/issues/5859

iaguis:
As far as I can see, systemd-machined only grows the filesystem at image import/pull time (https://github.com/systemd/systemd/search?utf8=%E2%9C%93&q=grow_machine_directory&type=).

PR #70, opened 2017-07-17
kube-spawn: enlarge storage pool for each cloned image

dongsupark:
Enlarge the storage pool that is actually a btrfs volume backed by `/var/lib/machines.raw`. Every btrfs volume would need more space for its metadata as much as the volume grows through CoW operations. `/var/lib/machines` should have more space than its original image size. Otherwise `/var/lib/machines` could become full when files start to be written to nspawn containers.

In short, it's equivalent to the following command sequence:
<pre>
# umount /var/lib/machines
# qemu-img resize -f raw /var/lib/machines.raw POOLSIZE
# mount -t btrfs -o loop /var/lib/machines.raw /var/lib/machines
# btrfs filesystem resize max /var/lib/machines
# btrfs quota disable /var/lib/machines
</pre>

Unfortunately it's a temporary workaround. How much space we should add more to the image might depend on estimation during run-time operations. In the long run, systemd itself should be able to reserve more space for the storage pool, every time when it pulls an image to store in the pool.

Fixes https://github.com/kinvolk/kube-spawn/issues/66

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,7 +6,7 @@ ENV["LC_ALL"] = "en_US.UTF-8"
 
 Vagrant.configure("2") do |config|
   config.vm.box = "jhcook/fedora26"
-  config.vm.provision "shell", inline: "dnf install -y btrfs-progs docker git go kubernetes strace tmux"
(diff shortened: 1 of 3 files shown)


dongsupark:
I'm going to merge it, as this fix is necessary for testing other parts. If any problem happens regarding storage pools, feel free to file issues.

Issue #69, opened 2017-07-17
kube-spawn should exit with an error when overlay module is not loaded


blixtra:
Right now it only prints a warning. Then it hangs because the overlay module was not loaded. Let's just bail in this case.

blixtra:
Why not just run `modprobe overlay` ourselves when we detect this?

PR #72, opened 2017-07-18
kube-spawn: ensure that system requirements are satisfied

dongsupark:
Ensure that the system requirements are satisfied for starting `kube-spawn`. It's just like running the commands below:

<pre>
# modprobe overlay
# modprobe nf_conntrack
# echo "131072" > /sys/module/nf_conntrack/parameters/hashsize
</pre>

It was originally done by `scripts/vagrant-mod-env.sh`. However it's actually hard for end-users to take care of the requirements prior to running `kube-spawn`. So allow `kube-spawn` to do the initialization by itself a bit actively.

Also remove unnecessary warnings `"exit status 1"` while waiting on completion.

Fixes https://github.com/kinvolk/kube-spawn/issues/69

### DIFF
--- a/cmd/kube-spawn/kube-spawn.go
+++ b/cmd/kube-spawn/kube-spawn.go
@@ -17,12 +17,9 @@ limitations under the License.
 package main
 
 import (
-	"bufio"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"os"
-	"strconv"
 	"time"
 
 	"github.com/Masterminds/semver"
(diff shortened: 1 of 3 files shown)


robertgzr:
What do you think about moving this new code related to checking before startup out of `cmd/kube-spawn` making it part of `pkg/bootstrap`.
It's similar to `kubeadm`'s preflight checks... Would it be useful to be able to skip them?

robertgzr:
@dongsupark ^

dongsupark:
@robertgzr SGTM. I'll try it later.

dongsupark:
Rebased on master, and moved the relevant parts to `pkg/bootstrap/node.go`.
Also removed parts for disabling SELinux & firewalld from the PR, as it could cause side-effects. (as @schu mentioned) Created [another issue](https://github.com/kinvolk/kube-spawn/issues/76) to address that part.

I didn't add the `--skip-preflight-checks` option to kubeadm, because then it would skip [a lot of checks](https://github.com/kubernetes/kubernetes/blob/release-1.7/cmd/kubeadm/app/preflight/checks.go#L488..L512). I'm not sure we can skip all of them.

PR #74, opened 2017-07-18
Vagrantfile: support multiple machines

dongsupark:
Support both Fedora 26, Ubuntu 17.04 (Zesty), and Debian testing. It defaults to Fedora like before.
Also update `vagrant-mod-env.sh` to support both usernames, `vagrant` and `ubuntu`.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,7 +19,7 @@ Vagrant.configure("2") do |config|
   end
 
   # Fedora 26
-  config.vm.define "fedora" do |fedora|
+  config.vm.define "fedora", primary: true do |fedora|
     config.vm.provision "shell", inline: "dnf install -y btrfs-progs docker git go kubernetes qemu-img strace tmux"
 
     config.vm.synced_folder ".", "/vagrant", disabled: true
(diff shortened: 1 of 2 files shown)


dongsupark:
Force-pushed with Debian support, as well as necessary changes in the mod script.

schu:
>  It defaults to Fedora like before.

It starts all three VMs by default. For Fedora only, it would need something like the following change, I think.

<details>

```
diff --git a/Vagrantfile b/Vagrantfile
index 757c711..26198fa 100644
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,7 +19,7 @@ Vagrant.configure("2") do |config|
   end
 
   # Fedora 26
-  config.vm.define "fedora" do |fedora|
+  config.vm.define "fedora", primary: true do |fedora|
     config.vm.provision "shell", inline: "dnf install -y btrfs-progs docker git go kubernetes qemu-img strace tmux"
 
     config.vm.synced_folder ".", "/vagrant", disabled: true
@@ -38,7 +38,7 @@ Vagrant.configure("2") do |config|
   end
 
   # Ubuntu 17.04 (Zesty)
-  config.vm.define "ubuntu" do |ubuntu|
+  config.vm.define "ubuntu", autostart: false do |ubuntu|
     config.vm.box = "ubuntu/zesty64"
     config.vm.provision "shell", inline: "curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -; echo \"deb http://apt.kubernetes.io/ kubernetes-xenial main\" > /etc/apt/sources.list.d/kubernetes.list; apt-get update; DEBIAN_FRONTEND=noninteractive apt-get install -y docker.io golang git qemu-utils selinux-utils systemd-container kubectl tmux"
 
@@ -55,7 +55,7 @@ Vagrant.configure("2") do |config|
   end
 
   # Debian testing
-  config.vm.define "debian" do |debian|
+  config.vm.define "debian", autostart: false do |debian|
     config.vm.box = "debian/testing64"
     config.vm.provision "shell", inline: "echo deb http://apt.dockerproject.org/repo debian-stretch main > /etc/apt/sources.list.d/docker.list; apt-get update; DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-unauthenticated docker-engine; DEBIAN_FRONTEND=noninteractive apt-get install -y golang git kubernetes-client qemu-utils selinux-utils systemd-container tmux"

```

Not the nicest user experience though when not using Fedora (since users then have to do e.g. `vagrant ssh debian`). Not sure it can be done better.
</details>

dongsupark:
@schu Fixed the autostart bug. Thanks!

Issue #75, opened 2017-07-19
systemd-networkd-wait-online times out


iaguis:
When booting the nodes, [`systemd-networkd-wait-online](https://www.freedesktop.org/software/systemd/man/systemd-networkd-wait-online.service.html) is started and keeps waiting until it times out (2 minutes). This means we add a 2 minute delay to every node:

```
Jul 19 10:03:41 kube-spawn-0 systemd-udevd[281]: Process '/usr/lib/systemd/systemd-vconsole-setup' failed with exit code 1.
Jul 19 10:05:29 kube-spawn-0 systemd-networkd-wait-online[111]: Event loop failed: Connection timed out
Jul 19 10:05:29 kube-spawn-0 systemd[1]: systemd-networkd-wait-online.service: Main process exited, code=exited, status=1/FAILURE
Jul 19 10:05:29 kube-spawn-0 systemd[1]: Failed to start Wait for Network to be Configured.
Jul 19 10:05:29 kube-spawn-0 systemd[1]: systemd-networkd-wait-online.service: Unit entered failed state.
Jul 19 10:05:29 kube-spawn-0 systemd[1]: systemd-networkd-wait-online.service: Failed with result 'exit-code'.
Jul 19 10:05:29 kube-spawn-0 systemd[1]: Reached target Network is Online.
Jul 19 10:05:29 kube-spawn-0 systemd[1]: Starting Docker Application Container Engine...
```

Note the delay between the first line and the second, and how Docker doesn't start until we reach target "Network is Online".

Even after that. `eth0` is in `configuring` state:

```
kube-spawn-0 ~ # networkctl -a
IDX LINK             TYPE               OPERATIONAL SETUP     
  1 lo               loopback           carrier     configured
  3 eth0             ether              routable    configuring
  4 docker0          ether              no-carrier  configuring

3 links listed.
```

We should figure out why this happens or disable/change the timeout of the unit since things seem to work anyway.

dongsupark:
I actually tried to override `docker.service` to make it decoupled with `network-online.target`. That's not simple, because it's not possible to change dependencies in drop-in unit files.

So I decided to simply mask `systemd-networkd-wait-online.service`. Please see https://github.com/kinvolk/kube-spawn/pull/78.

PR #78, opened 2017-07-19
scripts: mask systemd-networkd-wait-online.service

dongsupark:
Every time when `docker.service` runs, it waits for 120 seconds doing nothing, until `systemd-networkd-wait-online.service` times out. As a result, kubeadm is being blocked until docker server became available: a huge delay for every nspawn container, although network is already online and reachable.

Let's just mask `systemd-networkd-wait-online.service`, to allow `docker.service` to run without being blocked.

Fixes https://github.com/kinvolk/kube-spawn/issues/75

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -7,3 +7,6 @@ echo "root:k8s" | chpasswd
 systemctl enable docker.service
 systemctl enable kubelet.service
 systemctl enable sshd.service
+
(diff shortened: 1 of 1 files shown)


dongsupark:
Also added a line to set password for non-root user "core".

dongsupark:
I'll merge it as it's simple but necessary bugfixes.

PR #79, opened 2017-07-21
pkg/bootstrap: improve mechanism of GetRunningNodes()

dongsupark:
So far only systemd v232 or newer has been supported. One of the reasons is `GetRunningNodes()` being unable to parse output of `"machinectl list"`.

For example, for systemd v232 or newer:

<pre>
kube-spawn-0 container systemd-nspawn coreos 1478.0.0 10.22.0.130...
</pre>

OTOH systemd v231 or older:

<pre>
kube-spawn-0 container systemd-nspawn
</pre>

Let's improve parsing mechanism of `GetRunningNodes()` to support the legacy format of systemd v231 or older.

### DIFF
--- a/pkg/bootstrap/node.go
+++ b/pkg/bootstrap/node.go
@@ -82,7 +82,12 @@ func MachineImageExists(machine string) bool {
 func GetRunningNodes() ([]Node, error) {
 	var nodes []Node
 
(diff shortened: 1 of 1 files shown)


dongsupark:
Rebased and tested it again. It works fine. I'll merge.

Issue #54, opened 2017-07-11
Add start/stop commands


robertgzr:
To make the startup process simple and similar to `minikube` we should provide additional commands:

- [x] rename up -> setup
- [x] `start` wraps `machinectl pull  ... coreos`, `setup` and `init`
- [x] `stop` - does what `machinectl stop kube-spawn-{0,...,n}` would achieve. 

Retrieving the list of running nodes would rely on parsing `machinectl` output again (see #37)

dongsupark:
I'd like to say `"start"` instead of `"up"`, and `"stop"` instead of `"down"`. That way it sounds less ambiguous to me.
Though I'm open to discussion about the naming. Comments are welcome!

Issue #73, opened 2017-07-18
Separate files in cmd/


robertgzr:
With the additions in #72  https://github.com/kinvolk/kube-spawn/blob/dongsu/enlarge-storage-pool/cmd/kube-spawn/kube-spawn.go is already ~500 lines long.

In the future we want to add even more commands:
#38 #54 

We should split it into multiple files.

robertgzr:
Covered by #80

Issue #38, opened 2017-06-15
Add destroy command


robertgzr:
Currently you can take down the cluster manually by running:
```
machinectl stop kubeadm-nspawn-{0,1,...}
machinectl remove kubeadm-nspawn-{0,1,...}
```
Also: #11 #18 

We want to have one command that does this for us.

Depends on #37

robertgzr:
Replaced by #149

Issue #37, opened 2017-06-15
Store node names/IPs in ./.kube-spawn/


alban:
Currently, `kube-spawn` does not store any state (list of nodes with their names and IPs). It relies on `machined` to know what nodes are running. However, it causes some problems:
- a developer cannot run several tests in parallel: the names of the nodes will conflict
- to get the IPs, #36 parses the output of `machinectl list` but the format of output changes from one version to another.

I would store the state in `./.kube-spawn/` in the same way than Vagrant stores its state in `./.vagrant/`.

Then, the commands `kube-spawn init` and `kube-spawn destroy` would use the state in `.kube-spawn/` to know which nodes to init/delete.

robertgzr:
Replaced by #137

PR #80, opened 2017-07-21
kube-spawn: add start/stop commands

dongsupark:
* rename command from `up` to `setup`
* split out commands into multiple files to maintain the source tree better. `init.go` for init command, `setup.go` for setup command etc.
* introduce `start` and `stop` commands. `start` wraps `machinectl pull ... coreos`, `setup` and `init`. `stop` does what `machinectl stop kube-spawn-{0,...,n}` would achieve

Fixes https://github.com/kinvolk/kube-spawn/issues/54

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ $ make vendor all
 $ sudo machinectl pull-raw --verify=no https://alpha.release.core-os.net/amd64-usr/current/coreos_developer_container.bin.bz2 coreos
 
 # Spawn and provision nodes for the cluster
-$ sudo GOPATH=$GOPATH CNI_PATH=$GOPATH/bin ./kube-spawn up --image=coreos --nodes=3
+$ sudo GOPATH=$GOPATH CNI_PATH=$GOPATH/bin ./kube-spawn setup --image=coreos --nodes=3
 
 # Setup Kubernetes
 $ sudo GOPATH=$GOPATH CNI_PATH=$GOPATH/bin ./kube-spawn init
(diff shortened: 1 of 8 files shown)


robertgzr:
Also closes #73

robertgzr:
I see that the `stop` command also removes the images. Can't we have that as two separate commands? Like `stop` to just bring down the cluster and `destroy` to remove the images? ( #38 )

I would expect that if I run:
```
$ kube-spawn start
$ kubectl ...
$ kube-spawn stop
```
That I can just get my cluster back up by just running `kube-spawn start` again.
I guess that still needs #37 right?

dongsupark:
@robertgzr Thinking again about it closely, I think you're right. It makes sense to split it out into 2 separate commands: `kube-spawn stop` for `machinectl poweroff` (and/or `terminate`), and `kube-spawn destroy` for `machinectl remove`.
I'm going to exclude the latter from this PR, to create another PR to address #38.

As far as I see, without removing images, it would sometimes fail to restart a machine. To fix it, #37 needs to be implemented in the future. Users probably need to keep in mind the restriction.

dongsupark:
Updated.

* Excluded parts for removing images.
* Added an option `'--force'` to the `stop` command to run `machinectl terminate` instead of `poweroff`.
* Updated `scripts/vagrant-setup-env.sh` to replace `up` with `setup`.
* Let the `up` command to print out only a deprecation warning.

dongsupark:
Also updated README.

Issue #82, opened 2017-07-24
kube-spawn: up succeeds even when no node spawned


dongsupark:
'kube-spawn up' succeeds even when no node got spawned. Images are already there, but machines are not started.

```
$ sudo GOPATH=/home/vagrant/go CNI_PATH=/home/vagrant/go/bin ./kube-spawn up --nodes 2 --image coreos
2017/07/24 14:34:54 Checking base image
2017/07/24 14:34:54 k8s/kubelet already downloaded, skipping...
2017/07/24 14:34:54 k8s/kubeadm already downloaded, skipping...
2017/07/24 14:34:54 k8s/kubectl already downloaded, skipping...
2017/07/24 14:34:54 k8s/kubelet.service already downloaded, skipping...
2017/07/24 14:34:54 k8s/10-kubeadm.conf already downloaded, skipping...
Image resized.
Resize '/var/lib/machines' of 'max'
2017/07/24 14:34:55 All nodes are running. Use machinectl to login/stop/etc.
$ sudo machinectl list
No machines
```

PR #83, opened 2017-07-24
kube-spawn: fix bug when checking for a running node

dongsupark:
Only stopped nodes should be included in the list `nodesToRun`, which should be passed to iteration of `runNode()`. If that list is empty, just exit immediately printing out a warning.

Also rename `nodes` to `numNodes`, `NodeExists()` to `IsMachineImageExists()`,
`nodesToCreate` to `nodesToRun`.

Fixes https://github.com/kinvolk/kube-spawn/issues/82

### DIFF
--- a/pkg/bootstrap/node.go
+++ b/pkg/bootstrap/node.go
@@ -67,7 +67,7 @@ func NewNode(baseImage, machine string) error {
 	return nil
 }
 
-func NodeExists(machine string) bool {
+func MachineImageExists(machine string) bool {
 	// TODO: we could also parse machinectl list-images to find that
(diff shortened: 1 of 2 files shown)


pkg/bootstrap/node.go
```diff
@@ -67,7 +67,7 @@ func NewNode(baseImage, machine string) error {
 	return nil
 }
 
-func NodeExists(machine string) bool {
+func IsMachineImageExists(machine string) bool {
```
blixtra: Should this not be `MachineImageExists`? Don't get why we need to the `Is`. Doesn't read right.

dongsupark: @blixtra Sure, will change it to `MachineImageExists`.

Issue #81, opened 2017-07-24
kube-spawn: up command fails with "bridge: no such file or directory"


dongsupark:
Sometimes `kube-spawn up` fails like below.

```
$ sudo GOPATH=$GOPATH CNI_PATH=$GOPATH/bin ./kube-spawn up --image=coreos --nodes=3
2017/07/24 14:34:11 Checking base image
2017/07/24 14:34:11 k8s/kubelet already downloaded, skipping...
2017/07/24 14:34:11 k8s/kubeadm already downloaded, skipping...
2017/07/24 14:34:11 k8s/kubectl already downloaded, skipping...
2017/07/24 14:34:11 k8s/kubelet.service already downloaded, skipping...
2017/07/24 14:34:11 k8s/10-kubeadm.conf already downloaded, skipping...
Image resized.
Resize '/var/lib/machines' of 'max'
error: fork/exec bridge: no such file or directory
2017/07/24 14:34:11 unexpected result output: 
2017/07/24 14:34:11 Error running node: unable to parse result: unexpected end of JSON input
```

Kernel modules are loaded, SELinux & firewalld are disabled.
`bridge` binary is already installed.
CNI config already exists, env variable `CNI_PATH` is set to `$GOPATH/bin`.

I cannot reproduce it myself. So we should try to figure out how that could happen.

dongsupark:
This issue turned out to be actually `$GOPATH` being unset in the first place. I think we should just make `$GOPATH` default to `$HOME/go`. If that directory still doesn't exist, print out a big warning to users. Same thing for `$CNI_PATH` too.

robertgzr:
Isn't `$GOPATH` supposed to be set to `$HOME/go` by default since some recent version of Go:

> If no GOPATH is set, it is assumed to be $HOME/go on Unix systems
(https://github.com/golang/go/wiki/SettingGOPATH)

dongsupark:
@robertgzr Yes, since go 1.8. https://golang.org/doc/go1.8#gopath
Though for some reason it didn't work out as expected.
For now I would rather detect GOPATH and CNI_PATH in a flexible way. In the future we should remove unnecessary checks.

robertgzr:
btw you say this was partly covered by #84 
Wha is missing?

dongsupark:
@robertgzr I think we should somehow document these in requirements or troubleshooting. I'll do it.

PR #84, opened 2017-07-25
cnispawn, nspawntool: check GOPATH, CNI_PATH in a flexible way

dongsupark:
Check for availability of `$GOPATH` and `$CNI_PATH` in a flexible way. If a given path is not a valid path to a directory, fall back to the next possible path candidate. If it still fails, print out warning. That way, users could get chances to fix the paths.

Partly covers https://github.com/kinvolk/kube-spawn/issues/81

### DIFF
--- a/pkg/cnispawn/netns.go
+++ b/pkg/cnispawn/netns.go
@@ -26,11 +26,12 @@ import (
 
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/kinvolk/kube-spawn/pkg/bootstrap"
+	"github.com/kinvolk/kube-spawn/pkg/utils"
 )
 
 var (
-	gopath  string = os.Getenv("GOPATH")
-	cniPath string = os.Getenv("CNI_PATH")
+	goPath  string
+	cniPath string
 )
 
 type CniNetns struct {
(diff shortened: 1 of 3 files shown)


pkg/cnispawn/netns.go
```diff
@@ -19,25 +19,48 @@ package cnispawn
...
+	if !utils.IsValidDir(cniPath) {
+		// fall back to $GOPATH/bin
+		cniPathOrig := cniPath
+		cniPath = path.Join(goPath, "bin")
+		log.Printf("cnispawn: invalid CNI_PATH %q, fall back to %s...\n", cniPathOrig, cniPath)
+		if !utils.IsValidDir(cniPath) {
+			return nil, fmt.Errorf("cnispawn: invalid CNI_PATH %q", cniPath)
+		}
+	}
+
```
robertgzr: not sure but could all of this go into `pkg/util` so you have something like this?
```
goPath, err := util.ValidateGoPath(os.Getenv("GOPATH"))
if err ...
cniPath, err := util.ValidateCniPath(...)
```

Because it looks like we have things double...

dongsupark: @robertgzr Right. I did restructure the code to avoid duplication.

Issue #85, opened 2017-07-25
rootfs gets read-only mounted when spawning many nodes


dongsupark:
When spawning so many nodes that the host cannot normally handle, the host eventually could not find any more free space, and the rootfs (or storage volume for `/var/lib/machines`) gets read-only mounted. At that point, there's actually nothing users can do to recover from it. Remounting the volume as read-write doesn't always succeed.

I'm not sure about how to prevent it from happening. It might be worth warning users about possible storage explosion in the first place.

dongsupark:
One of the obvious things to do is improving [`bootstrap.EnlargeStoragePool()`](https://github.com/kinvolk/kube-spawn/blob/master/pkg/bootstrap/node.go#L113) so that it won't make the pool explode. To do that, we should check the current status of storage volume, get the amount of free storage space, and use that to set the maximum size of the storage pool.

PR #87, opened 2017-07-25
pkg/bootstrap: check if there's enough space in storage volume

dongsupark:
Before enlarging storage volume, we should first check if the additional volume size would make the whole storage volume full. If that might be the case, we should not enlarge the volume, and print warning to users.

Possibly resolves https://github.com/kinvolk/kube-spawn/issues/85

### DIFF
--- a/pkg/bootstrap/node.go
+++ b/pkg/bootstrap/node.go
@@ -214,7 +214,8 @@ func GetPoolSize(baseImage string, nodes int) (int64, error) {
 	// We should reserve some unallocated free space for the whole host,
 	// as 100% usage of rootfs could badly affect the system's reliability.
 	if extraSize >= int64((float64(freeVolSpace))*0.9) {
-		return 0, fmt.Errorf("no more free space in %s. Please get more space!", varDir)
+		biSizeMB := int64(biSize / 1024 / 1024)
+		return 0, fmt.Errorf("not enough space on disk for %d nodes, Each node needs about %d MB, so in total you'll need about %d MB available.", nodes, biSizeMB, int64(nodes)*biSizeMB)
 	}
 
 	poolSize += extraSize
--- a/pkg/bootstrap/node.go
+++ b/pkg/bootstrap/node.go
@@ -179,8 +179,7 @@ func IsNodeRunning(nodeName string) bool {
 }
 
 func GetPoolSize(baseImage string, nodes int) (int64, error) {
-	var poolSize int64  // in bytes
-	var extraSize int64 // in bytes
+	var poolSize, extraSize, biSize int64 // in bytes
 
 	// Give 50% more space for each cloned image.
 	// NOTE: this is just a workaround, as how much space we should add more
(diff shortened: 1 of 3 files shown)


dongsupark:
Rebased and added more comments. Tested again, it works.

blixtra:
I tried this with the following command...

`sudo GOPATH=$GOPATH CNI_PATH=$GOPATH/bin ./kube-spawn setup --image=coreos --nodes=20`

...knowing it would be too much for my VM.

```
2017/08/03 16:31:55 cannot get pool size: no more free space in /var/lib/. Please get more space!
2017/08/03 16:31:55 Error running node: unable to create overlayfs on "/home/chris/go/src/github.com/kinvolk/kube-spawn/.kube-spawn": overlayfs is not available. Try to pass a directory with a different filesystem (like ext4 or XFS) to --kube-spawn-dir.
```

The first error valid but the second is not The partition is `ext4`.

dongsupark:
@blixtra I think this issue is just the same as https://github.com/kinvolk/kube-spawn/pull/100.
It probably means that the kernel module "overlay" is not loaded.
So I've rebased this PR on top of master. Now it should be fine.

dongsupark:
And I've just found out another bug in this PR, about calculation of pool size. Pushed a fix.

blixtra:
I don't agree with the behaviour of this issue. Currently the check seems to just be a suggestion, the nodes are still created. I think it would be a much better user experience to do the calculation, and if we know that there is not enough space do nothing accept, report that too many nodes have been requested and print how many would be possible.

As a user, I'd rather not have to manually remove all the nodes.

dongsupark:
@blixtra Fixed it by doing cleanup images + hard exit in the case. Now users won't have to clean them up.

blixtra:
@dongsupark but why even create them if we know it will be too much? For example, I assume that the `.kube-spawn` file is still populated from this call. Ideally that should also remain untouched if we know we can't provide a cluster of the size requested.

dongsupark:
@blixtra I'm sure that each machine-specific directory `.kube-spawn/default/kube-spawn-*` will not be created with this PR included. If the unnecessary machine-specific directories are still there, that's probably just leftover from previous versions of kube-spawn. 
In the long run, [`kube-spawn destroy` command](https://github.com/kinvolk/kube-spawn/issues/38) will be able to clean up any machine-specific `.kube-spawn/default/kube-spawn-*` directories, as well as all other files like CNI configs: TODO

On the other hand, I agree that ideally this part should be implemented somehow differently, like getting the number of to-be-created-nodes before running `bootstrap.NewNode()`. Doing that, we would not have to clean up images afterwards. However, currently `bootstrap.NewNode()` relies on only the result of `machinectl clone`, and `machinectl clone` has no `--dry-run` option that just tells whether the machine can be cloned or not, instead of actually cloning the machine. So at the moment running actually `machinectl clone` is the only way to say that the machine is to be included in `[]nodesToRun`. Right now I just don't have a better idea.

dongsupark:
Per offline discussions, changed it to a rough estimation before getting list of the nodes. If it fails at `bootstrap.GetPoolSize()`, just fail hard with a friendly error message, without running any node.

pkg/bootstrap/node.go
```diff
@@ -188,35 +188,54 @@ func EnlargeStoragePool(baseImage string, nodes int) error {
...
 	}
 
-	poolSize += int64(float64(fiBase.Size())*extraSizeRatio) * int64(nodes)
+	// extraSize, space to be allocated, shoud be 90% of freeVolSpace,
+	// actual free space on the target volume. Here 90% is simply a
+	// pre-defined estimation of how much space can be occupied.
+	// We should reserve some unallocated free space for the whole host,
+	// as 100% usage of rootfs could badly affect the system's reliability.
+	if extraSize >= int64((float64(freeVolSpace))*0.9) {
+		return 0, fmt.Errorf("no more free space in %s. Please get more space for creating %d nodes, each of which needs space of %d bytes.", varDir, nodes, biSize)
```
blixtra: This msg is not really accurate. There is free space because we haven't filled it with images because we detected that it would have taken all the free space. Let me suggest... "not enough space on disk for %d nodes. Each node needs about %d MB, so in total you'll need about %d MB available."

dongsupark: @blixtra Fixed the message.

Issue #88, opened 2017-07-26
mount host directories into /var/lib/docker inside containers


dongsupark:
We have further storage problems inside nspawn containers, especially in `/var/lib/docker`, e.g. when running microservices like sock shop. To address the issue, we should mount a container-dedicated directory on the host, into `/var/lib/docker` on the container.

For example, we mount `.kube-spawn/kube-spawn-0` into `/var/lib/docker` on `kube-spawn-0`, `.kube-spawn/kube-spawn-1` into `/var/lib/docker` on `kube-spawn-1`, etc.

Then the docker volumes can be handled from rootfs on the host, and we would avoid most of storage issues.

Downside with the scenario is that rootfs on the host must be in filesystem that supports overlayfs. Ext4 does support overlayfs, while xfs, btrfs don't. 

Suggested by @iaguis

blixtra:
Can we have the mount directories in a subdirectory? I wouldn't like to clutter the top level config directory with potentially many directories. I'd suggest `mount`.

dongsupark:
@blixtra Ok, will do, just like `.kube-spawn/mount/kube-spawn-0`.

robertgzr:
@alban suggested in #37 that we should maybe have some idea namespaces. So you can maybe have multiple clusters/test runs without having the node names collide.

My suggestion:
```
.kube-spawn/
├── cluster-0
│   ├── meta
│   ├── node-0
│   │   ├── meta
│   │   └── mount
│   └── node-1
└── cluster-1
```

dongsupark:
@robertgzr That directory structure looks really nice. We should definitely follow such a general structure.
But what does `"cluster"` mean here? Was there any previous discussion?

blixtra:
@dongsupark I think that would be the name of the cluster. Would be user defined but unique in the system.

robertgzr:
@dongsupark 👉  #37 
The idea was that when you run `kube-spawn` multiple times without clean up you will have multiple occurrences of e.g. `kube-spawn-0`

I don't think we need to implement that now just that we should think ahead. Maybe we can put everything in `.kube-spawn/default` for now.

blixtra:
We could have one called `default` for now and use that when we actually implement the concept and the user doesn't specify.

dongsupark:
ok, will do with `.kube-spawn/default/kube-spawn-0/mount`, etc.

PR #89, opened 2017-07-26
bootstrap, nspawntool: mount .kube-spawn into /var/lib/docker

dongsupark:
Mount `.kube-spawn/default/MACHINE_NAME/mount` into `/var/lib/docker` on `MACHINE_NAME`. For example, `.kube-spawn/default/kube-spawn-0/mount` into `/var/lib/docker` on `kube-spawn-0`. Then the docker volume can be handled from rootfs on the host, and we could avoid most of storage issues.

Downside with the scenario is that rootfs on the host must be a filesystem that supports overlayfs. For example ext4 does support overlayfs, while others don't. If the host rootfs is non-supported filesystem, print out error and exit, so that users can run `"setup"` with an option `"--kube-spawn-dir=OTHER_DIR"`, where `OTHER_DIR` is ext4.

Fixes https://github.com/kinvolk/kube-spawn/issues/88

### DIFF
--- a/cmd/kube-spawn/setup.go
+++ b/cmd/kube-spawn/setup.go
@@ -28,8 +28,9 @@ import (
 )
 
 var (
-	numNodes  int
-	baseImage string
+	numNodes     int
+	baseImage    string
+	kubeSpawnDir string
 
 	cmdSetup = &cobra.Command{
 		Use:   "setup",
@@ -43,6 +44,7 @@ func init() {
 
 	cmdSetup.Flags().IntVarP(&numNodes, "nodes", "n", 1, "number of nodes to spawn")
 	cmdSetup.Flags().StringVarP(&baseImage, "image", "i", "", "base image for nodes")
+	cmdSetup.Flags().StringVarP(&kubeSpawnDir, "kube-spawn-dir", "d", "", "path to .kube-spawn directory")
 }
 
 func checkK8sStableRelease(k8srel string) bool {
@@ -127,7 +129,7 @@ func doSetup(numNodes int, baseImage string) {
 	}
 
 	for _, name := range nodesToRun {
-		if err := nspawntool.RunNode(k8srelease, name); err != nil {
+		if err := nspawntool.RunNode(k8srelease, name, kubeSpawnDir); err != nil {
 			log.Fatalf("Error running node: %s", err)
 		}
 
--- a/pkg/nspawntool/run.go
+++ b/pkg/nspawntool/run.go
@@ -30,6 +30,7 @@ import (
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 	cniversion "github.com/containernetworking/cni/pkg/version"
 
+	"github.com/kinvolk/kube-spawn/pkg/bootstrap"
 	"github.com/kinvolk/kube-spawn/pkg/utils"
 )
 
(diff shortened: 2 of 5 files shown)


pkg/nspawntool/run.go
```diff
@@ -83,6 +84,23 @@ func RunNode(k8srelease, name string) error {
...
 	defaultBinds = getDefaultBinds(cniPath)
 
+	if kubeSpawnDir == "" {
+		kubeSpawnDir = parseBind("$PWD/.kube-spawn")
+	} else if err := utils.CheckValidDir(kubeSpawnDir); err != nil {
+		kubeSpawnDir = parseBind("$PWD/.kube-spawn")
+	}
+
+	if err := bootstrap.PathSupportsOverlay(kubeSpawnDir); err != nil {
+		return fmt.Errorf("unable to create overlayfs on %q, try another filesystem like ext4.", kubeSpawnDir)
```
iaguis: "try to pass a directory with a different filesystem (like ext4 or XFS) to --kube-spawn-dir."?

dongsupark: @iaguis Fixed. Thanks!

pkg/nspawntool/run.go
```diff
@@ -83,6 +84,23 @@ func RunNode(k8srelease, name string) error {
...
 	defaultBinds = getDefaultBinds(cniPath)
 
+	if kubeSpawnDir == "" {
+		kubeSpawnDir = parseBind("$PWD/.kube-spawn")
+	} else if err := utils.CheckValidDir(kubeSpawnDir); err != nil {
+		kubeSpawnDir = parseBind("$PWD/.kube-spawn")
+	}
+
+	if err := bootstrap.PathSupportsOverlay(kubeSpawnDir); err != nil {
+		return fmt.Errorf("try to pass a directory with a different filesystem (like ext4 or XFS) to --kube-spawn-dir.")
```
iaguis: Sorry, I ninja-edited the comment before and I guess you read the old version. I'd rather get a bit more context by getting the error from `PathSupportsOverlay()`:

```
return fmt.Errorf("unable to create overlayfs on %q: %v. Try to pass a directory with a different filesystem (like ext4 or XFS) to --kube-spawn-dir.
```

cmd/kube-spawn/setup.go
```diff
@@ -43,6 +44,7 @@ func init() {
 
 	cmdSetup.Flags().IntVarP(&numNodes, "nodes", "n", 1, "number of nodes to spawn")
 	cmdSetup.Flags().StringVarP(&baseImage, "image", "i", "", "base image for nodes")
+	cmdSetup.Flags().StringVarP(&kubeSpawnDir, "kube-spawn-dir", "k", "", "path to .kube-spawn directory")
```
robertgzr: We already use `-k`  for `--kubernetes-version` [here](https://github.com/kinvolk/kube-spawn/blob/master/cmd/kube-spawn/kube-spawn.go#L53)

pkg/nspawntool/run.go
dongsupark: @iaguis Done.

cmd/kube-spawn/setup.go
dongsupark: @robertgzr Right. I changed it to `-d`.

PR #93, opened 2017-07-26
Install all CNI plugins

### LINKED ISSUE #64: Install all cni plugins
We're currently only document `bridge` and `host-local` plugins although the `loopback` plugin is fetched in `vagrant-setup-env.sh`.

We should probably install all cni plugins by doing:

```
go get -u github.com/containernetworking/plugins/plugins/...
```

robertgzr:
Closes #64


robertgzr:
I would've put the 3 form `vagrant-setup-env.sh`. But it's easier to just have one line...

/cc @iaguis @alban

dongsupark:
As it won't cause any regression, I'll just merge.

PR #96, opened 2017-07-26
doc: describe more about troubleshooting tips

dongsupark:
Describe more about troubleshooting tips, mainly about setting up nodes as well as requirements. Fix typos and change names too.

Closes https://github.com/kinvolk/kube-spawn/issues/81

PR #102, opened 2017-07-28
pkg/nspawntool: move tmp directory into .kube-spawn

dongsupark:
Now that kube-spawn automatically creates `.kube-spawn` under `$PWD`, we should move `tmp` directory to `.kube-spawn/default`.

Change documents as well to recommend using `$KUBECONFIG` instead of long command-line options.

See also https://github.com/kinvolk/kube-spawn/pull/96#discussion_r129821464


dongsupark:
@blixtra Done. Now it's stored in `.kube-spawn/default/` directly.

PR #98, opened 2017-07-27
Revert "kube-spawn-runc: don't set Std{err,out,in}"

iaguis:
This reverts commit aa6486f8a03105bdf7c4de16de5d280ffd78ffe8.

It was preventing logs to work. I tested this and it works now.

PR #103, opened 2017-07-31
kube-spawn-runc: selectively pass Stdout/Stderr before running command

dongsupark:
Selectively pass Stdout/Stderr, by determining if they are terminal or not, with help of ioctl with `syscall.TCGETS`.

If we would always pass them, interactive mode of connection to containers will fail with error messages like `"container not started"`. If we would not pass them at all, then `"kubectl logs"` would not be able to print any logs.

Suggested by @iaguis
Supersedes https://github.com/kinvolk/kube-spawn/pull/98

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -21,6 +21,10 @@ import (
 	"log"
 	"os"
 	"path"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
 )
 
 var (
@@ -64,4 +68,11 @@ func GetValidCniPath(inGoPath string) (string, error) {
 	}
 
 	return cniPath, nil
+}
+
+// IsTerminal returns true if the given file descriptor is a terminal.
+func IsTerminal(fd uintptr) bool {
+	var termios syscall.Termios
+	_, _, err := unix.Syscall(unix.SYS_IOCTL, fd, uintptr(syscall.TCGETS), uintptr(unsafe.Pointer(&termios)))
+	return err == 0
 }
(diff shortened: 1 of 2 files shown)


pkg/utils/utils.go
```diff
@@ -21,6 +21,15 @@ import (
...
 	"os"
 	"path"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
+
+const (
+	getTermios = 0x5401 // syscall.TCGETS
```
iaguis: can't we just use `syscall.TCGETS` instead of defining these?

pkg/utils/utils.go
```diff
@@ -21,6 +21,15 @@ import (
...
+	setTermios = 0x5402 // syscall.TCSETS
```
iaguis: Actually, we don't need this, do we?

dongsupark: @iaguis Both are done. Changed the const to `syscall.TCGETS`, and removed `setTermios`.

pkg/utils/utils.go
```diff
@@ -21,8 +21,14 @@ import (
 	"log"
 	"os"
 	"path"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
 )
 
+const getTermios = syscall.TCGETS
```
iaguis: Do you define this to make the code easier to understand? I'd argue that for people not acquainted with terminal details, `getTermios` is not easier than `TCGETS` so we could just use `syscall.TCGETS` in the code below.

In any case, feel free to ignore this nit.

dongsupark: @iaguis Either way I don't care. I'll change it.

dongsupark:
@iaguis Done. Changed text and removed the variable `getTermios`.

Issue #105, opened 2017-08-02
Make socat available inside nspawn containers


dongsupark:
As `socat` is by default not available in Container Linux, apps like weave scope are not able to make use of port forwarding inside our nspawn containers.
We need to think about how to make the tool available inside nspawn containers. Add-on based on [torcx](https://github.com/coreos/torcx) could be one of the possibilities.

dongsupark:
torcx doesn't seem to work well with nspawn containers. Normally torcx wants `/etc/torcx/next-profile` to be specified for loading a user-defined profile. When the system then reboots, the next profile gets detected, tarball gets unpacked. But in nspawn containers, no matter how many times a container reboots, the next-profile never gets detected. Nothing but the predefined system-wide profile `docker` works for now. OTOH the next-profile mechanism works fine in a pure virtual machine.

It's also not trivial for us to define system-wide profile, because the docker profile is defined in coreos-overlay, which 3rd-party users won't be able to touch.

PR #106, opened 2017-08-03
kube-spawn, bootstrap: include static binary of socat in containers

dongsupark:
To be able to make kubelet's port-forward work, download static binary of `socat` to be installed under `/opt/bin/`.
This is workaround for now. In the future we need to find a proper solution for the issue.

Fixes https://github.com/kinvolk/kube-spawn/issues/105

### DIFF
--- a/cmd/kube-spawn/setup.go
+++ b/cmd/kube-spawn/setup.go
@@ -106,6 +106,18 @@ func doSetup(numNodes int, baseImage string) {
 		}
 	}
 
+	// NOTE: workaround for making kubelet work with port-forward.
+	// Ideally we should solve the port-forward issue by either
+	// creating general add-ons based on torcx, or creating our own
+	// container image, or at least building socat statically on our own.
+	ksExtraDir := ".kube-spawn/extras"
+	if err := os.MkdirAll(ksExtraDir, os.FileMode(0755)); err != nil {
+		log.Fatalf("Unable to create directory %q: %v.", ksExtraDir, err)
+	}
+	if err := bootstrap.DownloadSocatBin(ksExtraDir); err != nil {
+		log.Fatalf("Error downloading socat files: %s", err)
+	}
+
 	var nodesToRun []string
 
 	for i := 0; i < numNodes; i++ {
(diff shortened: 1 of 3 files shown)


alban:
Do we verify the GPG signatures of the downloaded binaries?

Or at least, check the sha512sum?

dongsupark:
@alban No we don't. That's definitely something we should do in the future, not only for socat but also for all k8s binaries.

cmd/kube-spawn/setup.go
```diff
@@ -106,6 +106,14 @@ func doSetup(numNodes int, baseImage string) {
 		}
 	}
 
+	// NOTE: workaround for making kubelet work with port-forward.
+	// Ideally we should solve the port-forward issue by either
+	// creating general add-ons based on torcx, or creating our own
+	// container image, or at least building socat statically on our own.
+	if err := bootstrap.DownloadSocatBin("./k8s"); err != nil {
```
iaguis: Should we put it in `.kube-spawn/extras/` or similar instead of `k8s/`?

iaguis:
> This is not enough to make it work. It seems /opt/bin is not in $PATH for the kubelet.

Doing

```
kube-spawn-1 ~ # ln -s /opt/bin/socat /usr/bin/socat
```

Fixes the issue.

cmd/kube-spawn/setup.go
dongsupark: @iaguis Interesting, is kubelet's $PATH different from the default one? I'll change `/opt/bin/socat` to `/usr/bin/socat`.
I'll put the binary into `.kube-spawn/extras` as well.

dongsupark:
@iaguis Done

pkg/bootstrap/kubernetes.go
```diff
@@ -73,3 +74,28 @@ func DownloadK8sBins(version, dir string) error {
...
 	return nil
 }
+
+func DownloadSocatBin(dir string) error {
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		os.MkdirAll(dir, os.ModePerm)
+	}
+
+	var fd *os.File
+	fpath := path.Join(dir, path.Base(staticSocatUrl))
```
iaguis: can you use `filepath.Join`?

dongsupark: @iaguis Done. I'll merge.

Issue #76, opened 2017-07-19
Allow to handle SELinux and firewalld during init phase


dongsupark:
During the initialization, kube-spawn automatically loads necessary kernel modules. But currently it does not disable SELinux enforcing mode and firewalld. When end-users don't do anything about that, kube-spawn would fail to run. A simple approach to solve it would be to disable SELinux enforcing mode (`setenforce 0`) and to stop firewalld (`systemctl stop firewalld.service`). However, that could cause other side-effects w.r.t system security.

Ideally kube-spawn should configure specific SELinux rules as well as firewall rules, instead of disabling SELinux and firewalld completely.

See also https://github.com/kinvolk/kube-spawn/pull/72

dongsupark:
I spent some time on this issue. Conclusion is:
* It's not possible to keep SELinux enforcing mode. There are too many different issues with SELinux. https://bugzilla.redhat.com/show_bug.cgi?id=1393371, https://bugzilla.redhat.com/show_bug.cgi?id=1462382, https://bugzilla.redhat.com/show_bug.cgi?id=1310464. I don't think these will be fixed in the near future. I also tried to set proper SELinux rules for writing data into `/var/lib/machines`, but `machinectl pull-raw` cannot move data anywhere when enforcing mode turned on. So let's keep running `"setenforce 0"` for now.
* As for firewalld, we don't need to stop firewalld, but just add a rule to allow CNI traffic by running iptables. See https://github.com/kinvolk/kube-spawn/pull/110

PR #110, opened 2017-08-04
pkg/bootstrap: allow CNI traffic, disable SELinux enforcing mode

dongsupark:
Instead of stop firewall, we need to just allow traffic through CNI network interface `cni0`.
Also set the default policy of `FORWARD` chain to `ACCEPT`.

Though as for SELinux, we should automatically turn off the enforcing mode, as there's no way around multiple issues.

Fixes https://github.com/kinvolk/kube-spawn/issues/76

### DIFF
--- a/doc/troubleshooting.md
+++ b/doc/troubleshooting.md
@@ -10,14 +10,12 @@ If any path-related problem still occurs, please try the following approaches:
 * try to run with `sudo -E kube-spawn ...` to pass normal-user's env variables
 * try to run with pre-defined `GOPATH` or `CNI_PATH`, for example `sudo GOPATH=/home/myuser/go CNI_PATH=/home/myuser/go/bin kube-spawn ...`
 
-## SELinux and firewalld
+## SELinux
 
-To run `kube-spawn`, it is recommended to turn off SELinux enforcing mode as well as to stop firewalld:
+To run `kube-spawn`, it is recommended to turn off SELinux enforcing mode:
 
 ```
 $ sudo setenforce 0
-$ sudo systemctl stop firewalld.service
-$ sudo iptables -P FORWARD ACCEPT
 ```
 
 However, it is also true that disabling security framework is not always desirable. So it is planned to handle security policy instead of disabling them. Until then, there's no easy way to get around.
--- a/pkg/bootstrap/node.go
+++ b/pkg/bootstrap/node.go
@@ -383,7 +383,8 @@ func EnsureRequirements() {
 
 	// insert an iptables rules to allow traffic through cni0
 	ensureIptables()
-	// TODO: handle SELinux
+	// check for SELinux enforcing mode
+	ensureSelinux()
 }
 
 func isOverlayfsAvailable() bool {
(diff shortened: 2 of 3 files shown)


pkg/bootstrap/node.go
```diff
@@ -506,3 +509,194 @@ func ensureConntrackHashsize() {
...
+	outstr, _ := cmdGet.Output()
+	sestatus := strings.TrimSpace(string(outstr))
+	if sestatus == "Enforcing" {
+		return true
+	}
+
+	return false
+}
+
+func disableSELinuxEnforcing() error {
```
blixtra: I thought we said we didn't want to do this without user intervention. Maybe @iaguis recalls.

iaguis: Right, that's a bit invasive so we should check for it and error out (not only warn) if it's enabled.

dongsupark: @blixtra Yes right.
That's why we wanted to find out a way to set SELinux policies with the enforcing mode turned on, when we created the issue #76. But as written in [the comment](https://github.com/kinvolk/kube-spawn/issues/76#issuecomment-320288720), I don't see a way to do that at the moment. In practice, the only way to make the whole init procedure work is to turn off enforcing mode by running `setenforce 0`. Then I think we should make `kube-spawn init` do that automatically, at least until SELinux issues became somehow resolved.

If you still think that user intervention is necessary for the SELinux mode, then I think that it also makes sense. It's matter of choice.

dongsupark: @iaguis All right. Then I'll make it error out if the enforcing mode is enabled.

dongsupark: @iaguis Done.

pkg/bootstrap/node.go
```diff
@@ -628,3 +629,41 @@ func ensureIptables() {
...
+	if sestatus == "Enforcing" {
+		return true
+	}
+
+	return false
+}
+
+func ensureSelinux() {
+	if isSELinuxEnforcing() {
+		log.Fatalln("ERROR: SELinux enforcing mode is enabled. exit.")
```
blixtra: Should we not guide the use as to what to do? For example, "SELinux enforcing mode is enabled. you will need to disable it with `sudo setenforce 0` for kube-spawn to work properly."

In general, let's always guide the user if we can.

dongsupark: @blixtra Done. Added the guide message.

doc/troubleshooting.md
```diff
@@ -10,14 +10,14 @@ If any path-related problem still occurs, please try the following approaches:
 * try to run with `sudo -E kube-spawn ...` to pass normal-user's env variables
 * try to run with pre-defined `GOPATH` or `CNI_PATH`, for example `sudo GOPATH=/home/myuser/go CNI_PATH=/home/myuser/go/bin kube-spawn ...`
 
-## SELinux and firewalld
```
iaguis: We're doing this programmatically now, should we still keep the iptables part of this section?

dongsupark: @iaguis Right. I'll remove the iptables part.

dongsupark: @iaguis Done. I'll merge it

Issue #104, opened 2017-07-31
kube-spawn-runc: only print old/new args when logging is enabled


iaguis:
We print the old and new arguments even if we don't enable logging with `KUBE_SPAWN_RUNC_LOG_PATH`. This is confusing for users because they'd see something like:

```
2017/07/31 16:05:34 old args: []string{"--log", "/run/docker/libcontainerd/containerd/000b8c5b6dfe994c518b3fc26cbcff3883a50d5e005c7736bba1f133ce3f1004/init/log.json", "--log-format", "json", "create", "--bundle", "/var/run/docker/libcontainerd/000b8c5b6dfe994c518b3fc26cbcff3883a50d5e005c7736bba1f133ce3f1004", "--console", "", "--pid-file", "/run/docker/libcontainerd/containerd/000b8c5b6dfe994c518b3fc26cbcff3883a50d5e005c7736bba1f133ce3f1004/init/pid", "000b8c5b6dfe994c518b3fc26cbcff3883a50d5e005c7736bba1f133ce3f1004"}
2017/07/31 16:05:34 new args: []string{"--log", "/run/docker/libcontainerd/containerd/000b8c5b6dfe994c518b3fc26cbcff3883a50d5e005c7736bba1f133ce3f1004/init/log.json", "--log-format", "json", "create", "--no-new-keyring", "--bundle", "/var/run/docker/libcontainerd/000b8c5b6dfe994c518b3fc26cbcff3883a50d5e005c7736bba1f133ce3f1004", "--console", "", "--pid-file", "/run/docker/libcontainerd/containerd/000b8c5b6dfe994c518b3fc26cbcff3883a50d5e005c7736bba1f133ce3f1004/init/pid", "000b8c5b6dfe994c518b3fc26cbcff3883a50d5e005c7736bba1f133ce3f1004"}
```

at the top of every pod's log.

We should only print those if we are logging.

Issue #111, opened 2017-08-07
Use the real kubeconfig path in $KUBECONFIG note


iaguis:
We say

```
Note: For kubectl to work, please set $KUBECONFIG to $HOME/go/src/github.com/kinvolk/kube-spawn/.kube-spawn/default/kubeconfig.
```

But my `$GOPATH` is in `$HOME/work/go`. Can we do better and give the actual path where `kubeconfig` is?

dongsupark:
Closed via https://github.com/kinvolk/kube-spawn/pull/112

PR #112, opened 2017-08-08
Fix message for kubeconfig, only log for runc when necessary

dongsupark:
Print args only when `logPath` (`$KUBE_SPAWN_RUNC_LOG_PATH`) is explicitly set. Otherwise the logs will appear at the top of every pod's log.

At the end of init command, print out message for setting `$KUBECONFIG` based on real lookups: first try `$PWD/.kube-spawn/default/kubeconfig`, then `$GOPATH/src/github.com/kinvolk/kube-spawn/.kube-spawn/default/kubeconfig`, then `$HOME/go/src/github.com/kinvolk/kube-spawn/.kube-spawn/default/kubeconfig`.

Fixes https://github.com/kinvolk/kube-spawn/issues/104 https://github.com/kinvolk/kube-spawn/issues/111

### DIFF
--- a/cmd/kube-spawn/init.go
+++ b/cmd/kube-spawn/init.go
@@ -25,6 +25,7 @@ import (
 	"github.com/kinvolk/kube-spawn/pkg/bootstrap"
 	"github.com/kinvolk/kube-spawn/pkg/distribution"
 	"github.com/kinvolk/kube-spawn/pkg/nspawntool"
+	"github.com/kinvolk/kube-spawn/pkg/utils"
 )
 
 const pushImageRetries int = 10
@@ -94,5 +95,5 @@ func doInit() {
 		}
 	}
 
-	log.Println("Note: For kubectl to work, please set $KUBECONFIG to $HOME/go/src/github.com/kinvolk/kube-spawn/.kube-spawn/default/kubeconfig.")
+	log.Println("Note: For kubectl to work, please set $KUBECONFIG to " + utils.GetValidKubeConfig())
 }
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -21,10 +21,16 @@ import (
 	"log"
 	"os"
 	"path"
+	"path/filepath"
 	"syscall"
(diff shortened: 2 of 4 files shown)


pkg/utils/utils.go
```diff
@@ -70,6 +86,26 @@ func GetValidCniPath(inGoPath string) (string, error) {
 	return cniPath, nil
 }
 
+func GetValidKubeConfig() string {
+	kcPath := filepath.Join(pwd, kcRelPath)
```
iaguis: For some reason this doesn't work for me. It prints:

```
2017/08/08 11:54:37 Note: For kubectl to work, please set $KUBECONFIG to .kube-spawn/default/kubeconfig
```

which is not good because setting `$KUBECONFIG` to that will only make kubectl work if I'm in the kube-spawn directory.

Can you use `os.Getwd()` instead?

cmd/kube-spawn/init.go
```diff
@@ -94,5 +95,5 @@ func doInit() {
 		}
 	}
 
-	log.Println("Note: For kubectl to work, please set $KUBECONFIG to $HOME/go/src/github.com/kinvolk/kube-spawn/.kube-spawn/default/kubeconfig.")
+	log.Printf("Note: For kubectl to work, please set $KUBECONFIG to %s\n", utils.GetValidKubeConfig())
```
iaguis: Why the change from Printf to Println?

dongsupark: Just a matter of taste, I just think that `Printf("This is %s\n", str)` look better than `Println("This is " + str)`. Doesn't matter actually. Will change it to Println.

iaguis: Oh I see, didn't realize there was an argument there. You can keep the `Printf`

pkg/utils/utils.go
dongsupark: @iaguis Updated. Now it uses `os.Getwd()` to get the current working directory, and prints out the whole abspath.

PR #113, opened 2017-08-08
Update readme with demo and use the `up` command

blixtra:
I've updated the README to include a running demo and to highlight the easy, one-liner `up` command over the 3 stage cluster bring-up.


dongsupark:
Looks good to me.

Issue #114, opened 2017-08-10
kube-spawn fails to check the version of the CoreOS Container Linux image


alban:
```
$ machinectl list-images 
NAME             TYPE RO  USAGE CREATED                      MODIFIED                    
coreos           raw  no  1.1G  Wed 2017-05-31 02:54:16 CEST Wed 2017-05-31 02:54:16 CEST
```
This image is too old and `kube-spawn` fails: it blocks on the following message forever:
```
[apiclient] Created API client, waiting for the control plane to become ready
```
I am trying `sudo machinectl remove coreos` and run the `kube-spawn up` command again and it works better (but still fails. I will report more issues)

dongsupark:
Yes, I see. `kube-spawn` needs to download Container Linux alpha (currently 1492, but I think beta 1465 is already fine). It doesn't check the version before, so `machinectl pull-raw` just skips downloading an image, when the old image already exists. It should be smarter.

Question is, how can we check the version? The base image `/var/lib/machines/coreos.raw` has no info about that. Before actually booting a machine, the coreos version is not visible.

dongsupark:
Ok, it's simple to check the version:
```
$ machinectl image-status coreos | grep OS\:
              OS: Container Linux by CoreOS 1478.0.0 (Ladybug)
```

BTW it's interesting, beta (1465) is not enough. It has to be alpha (1478 or newer). @.@
I'll create a PR for checking if the version is 1478 or newer.

PR #117, opened 2017-08-11
pkg/bootstrap: check for Container Linux version

dongsupark:
As kube-spawn needs Container Linux 1478.0.0 or newer, we need to check for the version.

If the version is too low, it fails and recommends users to remove the old local image, so that the next run of `kube-spawn` will download a new image automatically.

Fixes https://github.com/kinvolk/kube-spawn/issues/114

### DIFF
--- a/pkg/bootstrap/node.go
+++ b/pkg/bootstrap/node.go
@@ -28,6 +28,8 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+
+	"github.com/Masterminds/semver"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/bootstrap/node.go
```diff
@@ -687,3 +692,86 @@ func ensureSelinux() {
...
+		}
+	}
+
+	return nil
+}
+
+func ensureCoreosVersion() {
+	if err := checkCoreosVersion(); err != nil {
+		log.Println(err)
+		log.Fatalf("You will need to remove the image by 'sudo machinectl remove coreos' and download the version %s at least by 'sudo machinectl pull-raw --verify=no https://alpha.release.core-os.net/amd64-usr/current/coreos_developer_container.bin.bz2 coreos'.", coreosStableVersion)
```
alban: you could recommend the user to remove the old image and let `kube-spawn` downloads a new one automatically.

dongsupark: @alban Done. Now it recommends users to remove the old image, so that the next run of kube-spawn can download a new one automatically.

Issue #120, opened 2017-08-14
Check for source file of bind-mount in pkg/nspawntool


dongsupark:
Currently the general structure of `pkg/nspawntool` has rooms for improvement. One of the main issues is that there's no general way of checking if a certain source mount directory exists. If any source mount file does not exist, `systemd-nspawn` called with e.g. `--bind=INVALID_SRC_FILE` will fail silently without any decent error message. That sometimes makes debugging a lot harder than expected. It's also a good source of potential bugs, as any source file could be invalid in the future at any time.

To improve that, we should improve the structure of `pkg/nspawntool`:
* Define a list of pre-defined structure to include `--bind=`, source file, and destination file.
* Convert each entry of the list into a list of string that can be directly passed to `systemd-nspawn`.
  * During the conversion, if any source directory doesn't exist, error out immediately.

PR #121, opened 2017-08-14
pkg/nspawntool: rewrite nspawntool to validate srcdir

dongsupark:
Rewrite `RunNode()` in nspawntool to check errors in a structured way.
    
* Introduce a structure `bindOption` that represents a single bind option string, e.g. `--bind=/src/file:/dst/file`. Each instance has 3 entries, `bindPrefix`, `srcMount`, and `dstMount`, which would be `--bind=`, `/src/file`, and `/dst/file` repectively in the example.

* Initially a list of bind options is composed of type `[]bindOption`, which will be afterwards transformed into `[]string` that can be directly passed to systemd-nspawn. Those conversions are done by `buildDefaultBindsList()` as well as `buildK8sBindsList()`.

* For each `srcMount`, check if the path exists. If not, error out. This is necessary for avoiding such cases where systemd-nspawn gets stuck suddenly without printing any error, whenever a non-existent path is passed as `srcMount`.

Fixes https://github.com/kinvolk/kube-spawn/issues/120

### DIFF
--- a/pkg/nspawntool/run.go
+++ b/pkg/nspawntool/run.go
@@ -41,26 +41,161 @@ const bindrw string = "--bind="
 var (
 	goPath  string
 	cniPath string
+
+	k8sBinds     []string
+	defaultBinds []string
 )
 
-var k8sbinds []string
-var defaultBinds []string
+type bindOption struct {
+	bindPrefix string // either bindro or bindrw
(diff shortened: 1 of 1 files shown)


dongsupark:
Rebased.

Issue #56, opened 2017-07-11
Add a smoke test script


schu:
The script, for instance `./tests/smoke.sh`, would

* do a clean kubeadm-nspawn Vagrant setup,
* use kubectl to verify the successful setup of k8s,
* do a deploy and
* query the deployed service to verify it's running.

robertgzr:
To check k8s setup?
```kubectl -n kube-system get nodes```

What can we deploy to test?
https://github.com/kubernetes/kubernetes/tree/master/examples/https-nginx ?

robertgzr:
Or maybe https://github.com/kubernetes/kubernetes/tree/master/examples/guestbook

schu:
> To check k8s setup?

Yes, a minimal script to test everything (that is expected to work) works as expected.

PR #124, opened 2017-08-28
[WIP] Initial implementation of smoke tests

dongsupark:
(still in work-in-progress. Please do not merge)

A simple implementation of smoke tests based on Vagrant.
    
Fixes https://github.com/kinvolk/kube-spawn/issues/56
Depends on https://github.com/kinvolk/kube-spawn/pull/117


dongsupark:
This test works well when we run `./scripts/test-smoke.sh` on our own. Next step would be to run the script automatically when each PR is made.

Unfortunately at the moment CI services like semaphoreCI or circleCI are based on Ubuntu 14.04, where neither vagrant nor VirtualBox is installed. So I wanted to somehow install vagrant & virtualbox from test-smoke.sh. It didn't work well, as there's no kernel tree inside chroot environment, which prevents vbox kernel modules from getting built. Anyway it's not a good idea at all to build kernel modules every time.

Next attempt was to make use of vagrant-libvirt, instead of virtualbox. Then we don't need to build vbox kernel modules at all. For that, I needed to support libvirt in Vagrantfile of kube-spawn. (See https://github.com/dongsupark/kube-spawn/tree/dongsu/smoke-test )
That works perfectly on most recent distros like Fedora 26, Ubuntu 17.04, but not on Ubuntu 14.04 of SemaphoreCI, where vagrant-libvirt package doesn't exist. Installing the package manually caused the whole dependency chain to be fetched. Not good at all.

So I would rather keep this PR based on virtualbox instead of libvirt. Maybe later when semaphoreCI supported Ubuntu 16.04, we will be able to make test-smoke.sh run integrated with semaphoreCI.

scripts/test-smoke.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# Script to test kube-spawn
+
+set -eux
+set -o pipefail
+
+CDIR=$(cd "$(dirname "$0")" && pwd)
+cd "$CDIR"
```
schu: Better with `pushd`:

```
pushd "$CDIR"
trap 'popd' EXIT
```

scripts/test-smoke.sh
```diff
@@ -0,0 +1,39 @@
...
+RESCODE=$?
+if [[ "${RESCODE}" -eq 0 ]]; then
+	RES="SUCCESS"
+else
+	RES="FAILURE"
+fi
+
+echo "Test result: ${RES}"
+
+vagrant halt fedora
```
schu: Could also go into trap.

dongsupark:
Updated.
* added tests for creating deployments, services etc.
* Fixed a name of Fedora Vagrant image.
* Used trap for cleaning up.

schu:
As far as I can see, it doesn't test if the pods/services are actually running and reachable over the network endpoints. IMHO we should test this.

Vagrantfile
```diff
@@ -5,7 +5,7 @@ ENV["TERM"] = "xterm-256color"
 ENV["LC_ALL"] = "en_US.UTF-8"
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "jhcook/fedora26" # defaults to fedora
+  config.vm.box = "fedora/26-cloud-base" # defaults to fedora
```
nhlfr: Rebase needed?

scripts/test-smoke.sh
```diff
@@ -0,0 +1,40 @@
...
+
+set -eux
+set -o pipefail
+
+CDIR=$(cd "$(dirname "$0")" && pwd)
+pushd "$CDIR"
+trap 'popd' EXIT
+
+if ! vagrant version > /dev/null 2>&1; then
+  echo "Please install vagrant first"
```
nhlfr: Indentaions in this script are inconsistent. Sometimes you are using two spaces, sometimes tabs. Please choose the one pattern - I prefer tabs, but whatever you prefer WFM.

dongsupark:
@nhlfr Also I would need to change commands according to the new design of v0.2.
I would also need to try to run this PR on semaphore CI, to check if the total download size still exceeds the limit from the provider.
Also need to try either libvirt or virtualbox, etc.

dongsupark:
Rebased.
Updated every necessary change according to the redesigned CLI from the current master branch.
Added a connectivity test that exposes a deployment, and checks if the cluster IP is reachable from a node.

Issue #126, opened 2017-09-04
kube-spawn doesn't work if kube-spawn-dir is in an eCryptfs


iaguis:
Testing kube-spawn on a system where `$HOME` is an eCryptfs we discovered that docker fails to start on the nodes.

The reason is that we the `overlay` driver can't be used on top of `ecryptfs` (although docker's error message was saying "unknown", PR with a fix here: https://github.com/moby/moby/pull/34721).

It seems eCryptfs passes our `PathSupportsOverlay()` test, so we should probably do one of these:

* add an explicit check there
* change the logic to a whitelist of known-to-be-working filesystems

dongsupark:
I'd prefer to add an explicit check for eCryptfs. Relying on a whitelist doesn't look like a long-term solution..

PR #129, opened 2017-09-05
pkg/bootstrap: check also for ecryptfs under overlayfs

dongsupark:
As moby does not allow ecryptfs to be used as underlying filesystem of overlayfs, we should also check for ecryptfs in `PathSupportsOverlay()` to prevent ecryptfs from being used.

Fixes https://github.com/kinvolk/kube-spawn/issues/126

### DIFF
--- a/pkg/bootstrap/util.go
+++ b/pkg/bootstrap/util.go
@@ -26,9 +26,11 @@ import (
 )
 
 const (
-	tmpDir      string = ".kube-spawn/default"
-	FsMagicAUFS        = 0x61756673 // https://goo.gl/CBwx43
-	FsMagicZFS         = 0x2FC12FC1 // https://goo.gl/xTvzO5
+	tmpDir string = ".kube-spawn/default"
+
+	FsMagicAUFS     = 0x61756673 // https://goo.gl/CBwx43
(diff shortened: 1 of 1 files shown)


jonboulle:
LGTM

robertgzr:
@dongsupark does this still work after #161 ?

`tmpDir` should be under `/var/lib/kube-spawn/<cluster>` now, no?

dongsupark:
@robertgzr This PR has nothing to do with `tmpDir`. The new code does not use `tmpDir` at all. I didn't intend to change the line, but apparently vim-go + gofmt automatically changed it when saving the file.
So we can merge this PR as-is.

Apart from that, later we should clean up the file, as `CreateSharedTmpDir()` still uses the old variable `tmpDir`.

schu:
> Apart from that, later we should clean up the file, as `CreateSharedTmpDir()` still uses the old variable `tmpDir`.

`CreateSharedTmpDir()` is unused in current master (b82cb524f5706fdd869241c788767e2eb27168df), probably it can be removed then.

dongsupark:
This PR was already approved, (thanks Jon!) and it fixes a real issue. So I'll merge it.
BTW I'm going to remove `CreateSharedTmpDir()` later.

PR #131, opened 2017-09-06
[WIP] Check signature of CoreOS image

nhlfr:
systemd-nspawn/machinectl doesn't have a functionality of checking downloaded images by GPG signatures. We need to download the image and validate it by ourselves and then import it by machinectl.


dongsupark:
Closed in favor of https://github.com/kinvolk/kube-spawn/pull/251

PR #132, opened 2017-09-08
Cleanup and update `Makefile` and `gitignore`

schu:
Brace expansion doesn't work with all shells.

`rootfs.tar.xz` etc. is from the past and not used anymore.


schu:
Please double-check if those indeed all are obsolete.

PR #134, opened 2017-09-11
vagrant: Add vagrant-libvirt support

nhlfr:
vagrant-libvirt plugin allows to run vagrant VMs by libvirt, using KVM as a hypervisor. That change also required to use Fedora box compatible both with virtualbox and libvirt providers.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -5,11 +5,15 @@ ENV["TERM"] = "xterm-256color"
 ENV["LC_ALL"] = "en_US.UTF-8"
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "jhcook/fedora26" # defaults to fedora
+  config.vm.box = "fedora/26-cloud-base" # defaults to fedora
 
   # common parts
   if Vagrant.has_plugin?("vagrant-vbguest")
     config.vbguest.auto_update = false
+  end
+  config.vm.provider :libvirt do |libvirt|
+    libvirt.cpus = 2
+    libvirt.memory = 4096
   end
   config.vm.provider :virtualbox do |vb|
     vb.check_guest_additions = false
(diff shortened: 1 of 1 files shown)


dongsupark:
Yes, I would really need this change. But we would need to first support both libvirt and vagrant for 3 distros, Fedora, Ubuntu, and Debian.
And I would test all combinations carefully. When I tried to do that, I had some issues w.r.t. syntax of Vagrantfile.

nhlfr:
Makes sense, I will try to find provider-agnostic images for all distros.

nhlfr:
@dongsupark done, all images introduced in this PR work both with virtualbox and libvirt.

schu:
Not introduced here, but the `apt.dockerproject.org` is deprecated AFAICT: https://docs.docker.com/engine/installation/linux/docker-ce/debian/

Do we actually need docker on the host or only in the nspawn containers? Is there a reason why we use `docker.io` on Ubuntu but vanilla Docker on Debian?

Also, I'm note sure if we should bother to provide three Vagrant environments (will those actually be properly tested/used?) or rather only Fedora?

(I'm also a bit wary of using unknown third party images, `generic/...`. But then, I have no idea if `fedora/boxes` is "official" or only sounds so... No way to tell for app.vagrantup.com images, afaik.)

Vagrantfile
```diff
@@ -40,26 +44,26 @@ Vagrant.configure("2") do |config|
 
   # Ubuntu 17.04 (Zesty)
   config.vm.define "ubuntu", autostart: false do |ubuntu|
-    config.vm.box = "ubuntu/zesty64"
+    config.vm.box = "generic/ubuntu1704"
     config.vm.provision "shell", inline: "curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -; echo \"deb http://apt.kubernetes.io/ kubernetes-xenial main\" > /etc/apt/sources.list.d/kubernetes.list; apt-get update; DEBIAN_FRONTEND=noninteractive apt-get install -y docker.io golang git qemu-utils selinux-utils systemd-container kubectl tmux"
 
     config.vm.synced_folder ".", "/vagrant", disabled: true
     config.vm.synced_folder ".", "/home/ubuntu/go/src/github.com/kinvolk/kube-spawn",
```
dongsupark: I like the change from `ubuntu` from `vagrant`.
But we also need to change `/home/ubuntu` to `/home/vagrant` here in synced_folder. Otherwise rsync could not work.

dongsupark:
Other than that it seems to work for both Fedora and Ubuntu, for both Virtualbox and libvirt.
I didn't test with Debian.

I think we need to keep Fedora and Ubuntu at least, to be able to test occasionally if kube-spawn works for the major distros. I don't care about Debian though. We can maybe remove Debian.

nhlfr:
@schu The problem with `generic` image is that Ubuntu doesn't provide official images for libvirt provider - that's the only reason why we use a third-party image.

We can drop Debian (or even Ubuntu), but can we do that in the other PR? In this one I would like to just get everything we had working on libvirt provider and it seems that we already achieved that goal.

dongsupark:
@nhlfr I agree. This PR looks good as it is. Let's merge it.

The `config.vm.synced_folder` looks also correct. Thanks! 👍

Issue #123, opened 2017-08-23
version mismatch of k8s apiserver and controller-manager


dongsupark:
When testing with k8s dev version (1.8), we're seeing a different version for `apiserver` and `controller-manager`:

```
$ $GOPATH/src/k8s.io/kubernetes/_output/dockerized/bin/linux/amd64/kube-apiserver --version
Kubernetes v1.8.0-alpha.2.2330+7e208befd197d0

kube-spawn-0 ~ # /proc/$(pidof kube-apiserver)/exe --version
Kubernetes v1.8.0-alpha.2
```

Even though the hyperkube image built has the right versions:

```
$ docker run -it 10.22.0.1:5000/hyperkube-amd64
# ./apiserver --version
2017-08-23 13:28:53.306040 I | proto: duplicate proto type registered: google.protobuf.Any
2017-08-23 13:28:53.306080 I | proto: duplicate proto type registered: google.protobuf.Duration
2017-08-23 13:28:53.306092 I | proto: duplicate proto type registered: google.protobuf.Timestamp
Kubernetes v1.8.0-alpha.2.2330+7e208befd197d0
```

It might be because [`KUBE_HYPERKUBE_IMAGE` is deprecated since 1.8.](https://github.com/kubernetes/kubernetes.github.io/blob/74fe1170ccbbd4cdf4f27fa9d048105ed2b3ed5b/docs/admin/kubeadm.md#environment-variables)

Reported by @iaguis from https://github.com/kinvolk/kube-spawn/pull/118#issuecomment-324334058

dongsupark:
Same (or related) issue, reported by @iaguis.

`./kube-spawn up --kubernetes-version=1.7.5` is broken because `latest` is hardcoded in `etc/kubeadm.yml` and we use `--config` *and* `--kubernetes-version` when calling kubeadm, which is not allowed now.

if we can specify the version in `kubeadm.yml` we can stop using `--kubernetes-version` and we can put a placeholder in `kubeadm.yml` that will change according to the kube-spawn version flag.

robertgzr:
This will most likely be fixed with the changes currently wip on [kube-spawn@robertgzr/rktlet](https://github.com/kinvolk/kube-spawn/tree/robertgzr/rktlet)

PR #136, opened 2017-09-15
*: Support for both container runtimes, docker and rkt

dongsupark:
Make kube-spawn support not only docker but also rkt. Users are able to run kube-spawn setup with an option `--container-runtime=rkt`.                                                                     

To do that, we change a way of writing most scripts and configs, i.e. generate them according to different k8s version or different container runtimes.

I've only tested with k8s 1.7.5, for both docker and rkt.

Based on patches written by @robertgzr & @iaguis 
Fixes https://github.com/kinvolk/kube-spawn/issues/123


robertgzr:
@dongsupark can you squash that fixup commit?

dongsupark:
@robertgzr Will do.

cmd/kube-spawn/init.go
```diff
@@ -79,10 +84,37 @@ func doInit() {
...
 	}
 
+	token, err := getToken()
+	if err != nil {
+		log.Fatalf("Error reading token: %v", err)
+	}
+
+	if k8sruntime == "rkt" {
+		kubeadmContainerRuntime = "rktlet"
+	} else {
```
nhlfr: Woudn't it be a little bit more correct if we will look for "docker" value explicitly? Like:

```
switch k8sruntime {
case "", "docker":
    kubeadmContainerRuntime = "docker"
case "rkt":
    kubeadmContainerRuntime = "rkt"
default:
    log.Fatalf("runtime %s is not supported", k8sruntime)
}
```

scripts/bootstrap.sh
```diff
@@ -7,6 +7,7 @@ echo "core:core" | chpasswd
 
 systemctl enable docker.service
 systemctl enable kubelet.service
+systemctl enable rktlet.service
```
nhlfr: Does it mean that we run rktlet unconditionally, even if we are using docker runtine? Or am I missing something?

pkg/script/kubeadm-join.go
```diff
@@ -0,0 +1,25 @@
+package script
+
+import "bytes"
+
+const kubeadmJoinTmpl string = `#!/bin/sh
```
nhlfr: I'm a little bit confused. Why you put one script there, in the code, while the another are still in separate files? Can't we template scripts which are in separate files?

dongsupark: @nhlfr Right. Scripts should be basically generated by `pkg/script` to be written to `/var/lib/kube-spawn/scripts`, and scripts under `./scripts` of the source tree, such as bootstrap, init or join, should be removed. They are just leftovers. I'll clean them up.

pkg/nspawntool/run.go
```diff
@@ -152,163 +167,162 @@ func getK8sBindOpts(k8srelease, goPath string) ([]bindOption, error) {
...
-			return nil, err
-		}
-		listBinds = append(listBinds, outOpt)
+
+	var runtimeDir string
+	switch n.Runtime {
+	case "docker":
+		runtimeDir = "/var/lib/docker"
+	case "rkt":
+		runtimeDir = "/var/lib/rkt"
```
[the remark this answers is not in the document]
dongsupark: @iaguis Good catch. That's probably why kube-spawn is not able to download large files. Every k8s pods are stored under `/var/lib/rktlet`, not `/var/lib/rkt`. Will fix.

dongsupark:
Updated.
* Fixed mount point for rktlet to `/var/lib/rktlet`.
* Added a cmdline option `--rktlet-bin-dir`
* Changed to checking for specific runtime name, `docker` or `rkt`, etc.
* Removed unnecessary scripts and configs from the source tree.

cmd/kube-spawn/setup.go
```diff
@@ -102,65 +112,230 @@ func doSetup(numNodes int, baseImage, kubeSpawnDir string) {
...
+		RktRuntime:      (k8sruntime == "rkt"),
+		RuntimeEndpoint: kubeadmRuntimeEndpoint,
+		RequestTimeout:  kubeadmRequestTimeout,
+	})
+	if outbufRuntime == nil {
+		stderr.Fatalf("Error generating kubeadm init script")
 	}
 
-	log.Printf("All nodes are running. Use machinectl to login/stop/etc.")
+	// NOTE: the 1st part "outbuf" and the 2nd part "outbufRuntime" is merged into
```
robertgzr: why do we merge the unit files here? cri-o and rktlet both use `--container-runtime=remote` and the endpoint option.
Doestn't it make more sense to keep them seperate?

dongsupark:
Updated w.r.t. talks with @robertgzr 
* Combined extra-args and runtime into a single file `kubeadm-extra-args.go`. As the 2 different KUBELET_EXTRA_ARGS definitions need to be combined into a single definition, it doesn't make sense to keep them separate.
* Determined the `failSwapOnArgs` argument before calling `GetKubeadmExtraArgs()`, to avoid tricky syntax in the template.
* Split into multiple commits.

cmd/kube-spawn/setup.go
dongsupark: @robertgzr As discussed, I also would like to keep them separate, but it was not that trivial. The thing is, 2 different definitions of `KUBELET_EXTRA_ARGS` need to be merged into 1 definition in the end. Otherwise kubelet won't be able to recognize all the options. So we decided to keep it simply in a single source file.

pkg/script/kubeadm-bootstrap.go
```diff
@@ -0,0 +1,27 @@
+package script
+
+import "bytes"
+
+const kubeadmBootstrapTmpl string = `#!/bin/sh
```
schu: Why `sh` and not `bash`?

pkg/script/kubeadm-bootstrap.go
```diff
@@ -0,0 +1,27 @@
...
+set -ex
+
+echo "root:k8s" | chpasswd
+echo "core:core" | chpasswd
+
+systemctl enable {{.K8sRuntime}}.service
+systemctl enable kubelet.service
+systemctl enable sshd.service
+
+# necessary to prevent docker from being blocked.
```
schu: Can you explain why / add a link?

pkg/script/script.go
```diff
@@ -0,0 +1,16 @@
...
+	"bytes"
+	"fmt"
+	"text/template"
+)
+
+func render(tmpl string, opts interface{}) *bytes.Buffer {
+	var buf = new(bytes.Buffer)
+	t := template.Must(template.New(fmt.Sprintf("%T", opts)).Parse(tmpl))
+	if err := t.Execute(buf, opts); err != nil {
+		return nil
```
schu: Why not report the error?

cmd/kube-spawn/kube-spawn.go
```diff
@@ -20,40 +20,81 @@ import (
...
+	cmdKubeSpawn.Flags().BoolVarP(&printVersion, "version", "V", false, "Output version information")
+	cmdKubeSpawn.PersistentFlags().StringVarP(&k8sruntime, "kubernetes-runtime", "r", defaultRuntime, "Runtime to use for the spawned cluster (docker or rkt)")
+	cmdKubeSpawn.PersistentFlags().StringVar(&rktBinDir, "rkt-bin-dir", "", "path to rkt binaries")
+	cmdKubeSpawn.PersistentFlags().StringVar(&rktletBinDir, "rktlet-bin-dir", "", "path to rktlet binaries")
 	cmdKubeSpawn.PersistentFlags().StringVarP(&k8srelease, "kubernetes-version", "k", k8sStableVersion, "Kubernetes version to spawn, \"\" or \"dev\" for self-building upstream K8s.")
+	cmdKubeSpawn.PersistentFlags().StringVarP(&kubeSpawnDir, "kube-spawn-dir", "d", kubeSpawnDirDefault, "path to kube-spawn asset directory")
+
+	cmdKubeSpawn.PersistentPreRun = func(cmd *cobra.Command, args []string) {
+		stderr = log.New(os.Stderr, fmt.Sprintf("[%v] ", cmd.Name()), log.Lshortfile)
+		stdout = log.New(os.Stdout, "", 0)
```
schu: What's the reasoning behind this?

cmd/kube-spawn/setup.go
```diff
@@ -102,65 +111,220 @@ func doSetup(numNodes int, baseImage, kubeSpawnDir string) {
...
+		kubeadmRuntimeEndpoint = "unix:///var/run/docker.sock"
+	case "rkt":
+		kubeadmCgroupDriver = "systemd"
+		kubeadmRuntimeEndpoint = "unix:///var/run/rktlet.sock"
+	default:
+		stderr.Fatalf("runtime %s is not supported", k8sruntime)
+	}
+
+	// K8s 1.8 or newer fails to run by default when swap is enabled.
+	// So we should disable the feature with an option "--fail-swap-on=false".
```
schu: Is disabling swap not an option?

cmd/kube-spawn/setup.go
```diff
@@ -102,65 +111,220 @@ func doSetup(numNodes int, baseImage, kubeSpawnDir string) {
...
-			log.Fatalf("Error running bootstrap script: %s", err)
+		if err := node.Bootstrap(); err != nil {
+			stderr.Fatalf("Error running bootstrap script: %v", err)
 		}
 
-		log.Printf("Success! %s started.", name)
+		stdout.Printf("Success! %s started.", node.Name)
+	}
+
+	stdout.Print("All nodes are running. Use machinectl to login/stop/etc.")
```
schu: Nit: "Use machinectl to login ..." is confusing as long as `machinectl login ...` doesn't work :)

Personally, I would prefer better documentation in the README (for example, explain common task such as up, start, stop, delete, etc).

pkg/utils/utils.go
```diff
@@ -135,3 +136,30 @@ func IsTerminal(fd uintptr) bool {
...
+	if err != nil {
+		return "", err
+	}
+	defer d.Close()
+
+	_, err = io.Copy(d, s)
+	return dst, err
+}
+
+func LookupPwd(inPath string) string {
```
schu: Why do we need this function?

cmd/kube-spawn/kube-spawn.go
robertgzr: Having info and error output rather than printing everything as a log with date/time

schu: I meant, why do you want some messages go to stdout instead of stderr? Why do messages going to stdout not have the `[cmd]` prefix? If it's just for the format, a `log.SetPrefix(fmt.Sprintf("[%v] ", cmd.Name()))` and `log.SetFlags(log.Lshortfile)` would do as well. What is with all the messages from `pkg/...`?

IMHO, if you want to have "better" logging, try to write down a design first, as far as possible something which suffices both current (again, what about all the code in `pkg`?) and future log usage. And use one of the established libraries in Go (or, if really necessary, build a custom logger pkg compatible with the standard library logger).

Right now it's a wild mix, even in the main package (`git grep 'log\.' cmd/kube-spawn`).

robertgzr: I agree 100%
I started working on the output stuff in a separate commit that should have been excluded from this PR in the first place...

robertgzr: I will open a PR for the logging and output thing.

robertgzr:
Closing in favor of #138

pkg/script/kubeadm-bootstrap.go
robertgzr: it was that way in the original script... should I change it?

PR #138, opened 2017-09-19
Support for rktlet

### LINKED ISSUE #123: version mismatch of k8s apiserver and controller-manager

robertgzr:
Make kube-spawn support rkt. Introduces a `--container-runtime=` flag.

To support multiple runtimes we now generate the relevant config files and scripts dynamically.

Tested with Kubernetes v1.7.5 for rkt and docker.

Replaces #136 
Fixes #123


dongsupark:
Both `vagrant-setup-env.sh` and `vagrant-mod-env.sh` are needed by Vagrantfile. Without them, `vagrant up` fails.

pkg/nspawntool/run.go
```diff
@@ -199,116 +208,127 @@ func getK8sBindOpts(k8srelease, goPath string) ([]bindOption, error) {
...
-			bindPrefix: bindrw,
-			srcMount:   mountDir,
-			dstMount:   "/var/lib/docker",
-		}
-		outOpt, err := bo.composeBindOption()
-		if err != nil {
-			return nil, err
-		}
-		listBinds = append(listBinds, outOpt)
+		return nil, fmt.Errorf("unable to create dir %q: %v", mountDir, err)
```
dongsupark: It must be `err != nil`. Otherwise kube-spawn exits here always, and doesn't run at all.

schu:
When testing this with @iaguis  (commit 64ffa8736d7897ee5e31edfb41496e0744dc9a55), I additionally had to

* `ln -sfT /etc/cni/net.d /etc/rkt/net.d` (as in https://github.com/kinvolk/kube-spawn/commit/1d31f30b310f8cb79e1fa083a2b5ced2e7690b24)
* `mkdir -p /usr/lib/rkt/plugins/ && ln -s /opt/cni/bin/ /usr/lib/rkt/plugins/net`
* and make sure the `weave-net` binaries are in my host `$GOPATH/bin` (which is bind mounted to `/opt/cni/bin/`)

to make it work (i.e. pods can be deployed and exposed).

Probably that should be done as part of this PR.

robertgzr:
Addressed @schu 's comments and reorganized parts of the code.
I changed parts that were rkt specific to support adding code to spawn a cluster with the `cri-o` runtime (only the latest commit contains the actual cri-o related additions).

My tests of running `cri-o` on kube-spawn failed because it has [runtime dependencies](https://github.com/kubernetes-incubator/cri-o#build-and-run-dependencies) that are quite challenging to work around on container linux.

dongsupark:
Ok, I pushed another commit for fixing syntax errors. Now it seems to work for both docker and rkt. I haven't tested it with cri-o.

For using rktlet we should run it like that:
```
sudo GOPATH=$GOPATH CNI_PATH=$GOPATH/bin \
 KUBE_SPAWN_RKT_BIN=$GOPATH/bin/rkt \
 KUBE_SPAWN_RKT_STAGE1_IMAGE=$GOPATH/bin/stage1-coreos.aci \
 KUBE_SPAWN_RKTLET_BIN=$GOPATH/src/github.com/kubernetes-incubator/rktlet/bin/rktlet \
 ./kube-spawn --kubernetes-version=1.7.5 --container-runtime=rkt up --nodes 2
```

PR #139, opened 2017-09-21
[WIP] pkg/nspawntool: mount /run for each machine

dongsupark:
(Work-In-Progress. Please do not merge)

To be able to run cri-tools from the host, we should mount from `/var/lib/kube-spawn/default/<machine_name>/run` on the host to `/run` inside each machine. The main reason is that cri-tools requires accessing to a unix socket for the target machine.

Depends on https://github.com/kinvolk/kube-spawn/pull/138


dongsupark:
No, it's not a good idea to run critest inside nspawn containers. critest should be able to make use of a unix socket, which only rktlet listens on.
So I'm closing it. Sorry for noise.

PR #151, opened 2017-09-28
Fix k8sGithubURL and kubeadm config file name

iaguis:
Its name was changed recently. Switch to a commit to avoid this kind of
issues in the future. Same for the socat repo.


iaguis:
I tested it from a clean `/var/lib/kube-spawn` and it worked.

PR #152, opened 2017-09-29
Fix bugs in setting up k8s parameters

dongsupark:
For the ContainerRuntime parameter of `bootstrap.sh`, we need to pass an explicit string, either `rktlet` or `docker`. Without this fix, bootstrap.sh will silently fail in case of rkt, because there's no `rkt.service` to enable. As a result, the rest of commands in bootstrap.sh will not run at all.

Also fix a bug to pass `--fail-swap-on=false` for k8s 1.8. The kubelet option `--fail-swap-on=false` is needed not only by `dev` version, but also by k8s release 1.8.0 or newer. Without the fix, k8s 1.8 will not run at all.

Reported by @iaguis

### DIFF
--- a/cmd/kube-spawn/setup.go
+++ b/cmd/kube-spawn/setup.go
@@ -85,6 +85,24 @@ func doCheckK8sStableRelease(k8srel string) {
 	}
 }
 
+// The kubelet option --fail-swap-on=false is needed for 2 cases:
+//  - "dev" version, which is a self-built K8s cluster
+//  - release "1.8.0" or newer
+func isFailSwapOnFalseNeeded(k8srel string) bool {
+	if utils.IsK8sDev(k8srel) {
+		return true
+	}
+
+	v, err := semver.NewVersion(k8srel)
+	if err != nil {
+		return false
+	}
+
+	c, _ := semver.NewConstraint(">=1.8.0")
+
+	return c.Check(v)
+}
+
 func runSetup(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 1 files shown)


cmd/kube-spawn/setup.go
```diff
@@ -191,8 +214,17 @@ func doSetup(numNodes int, baseImage, kubeSpawnDir string) {
 }
 
 func writeKubeadmBootstrapScript() {
+	switch k8sruntime {
+	case "rkt":
+		kubeadmContainerRuntime = "rktlet"
+	case "", "docker":
+		kubeadmContainerRuntime = "docker"
```
iaguis: we can just do `fallthrough` here.

cmd/kube-spawn/setup.go
```diff
@@ -85,6 +85,29 @@ func doCheckK8sStableRelease(k8srel string) {
...
+		return true
+	}
+
+	v, err := semver.NewVersion(k8srel)
+	if err != nil {
+		return false
+	}
+
+	k8sFailSwapOnVersion := "1.8.0"
+	c, err := semver.NewConstraint(">=" + k8sFailSwapOnVersion)
```
iaguis: Since this is just a static string, we can ignore the error. Also, we can just write `>= 1.8.0` here.

dongsupark:
@iaguis Done.

iaguis:
Note this won't make v1.8.0 clusters work because we're still using `10-kubeadm-pre-1.8.conf`. But we can fix that in a follow-up PR.

PR #155, opened 2017-10-03
vendor: check in vendor trees

dongsupark:
To be able to run smoke tests remotely on SemaphoreCI or else, we need to optimize build performance for each build. For that, we need to get the vendor tree included in the source tree. Otherwise running `dep ensure` for each build could bring negative impact on build performance.

NOTE: this PR might include unnecessary files, such as test files or documents. However, at the moment `dep` doesn't provide a native way of excluding such files. Thus I made them included for now.

Fixes https://github.com/kinvolk/kube-spawn/issues/153

PR #156, opened 2017-10-04
Specify `dep ensure` in README

zeenix:
Without `dep ensure`, the build fails so we should specify that in the
README.

This is in conflict with https://github.com/kinvolk/kube-spawn/pull/155/ . i-e if we don't end up merging that work, then I think we should have this in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ $ go get -d github.com/kinvolk/kube-spawn
 ```
 # Build the tool
 $ cd $GOPATH/src/github.com/kinvolk/kube-spawn
+$ dep ensure -v
 $ make vendor all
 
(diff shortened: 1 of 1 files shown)


dongsupark:
@nhlfr We already have an [`update-vendor` target](https://github.com/kinvolk/kube-spawn/blob/6c5b6475bc970f180e569e582ef6ccbd13d7d113/Makefile#L29..L30). Previously the target was named just `vendor`, but it was recently renamed. If you run `make update-vendor`, it will run `dep ensure`.

@zeenix As the PR https://github.com/kinvolk/kube-spawn/pull/155 is already merged, users don't need to do `dep ensure` any more. Instead in README.md, we should change the line below, from `make vendor all` to `make update-vendor all`.

PR #158, opened 2017-10-05
docs: Specify URL to raw YAML rather than webpage

zeenix:
Otherwise, you get a very unfriendly error:

error: error converting YAML to JSON: yaml: line 328: mapping values are not allowed in this context

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ To verify everything worked you can run:
 $ export KUBECONFIG=/var/lib/kube-spawn/default/kubeconfig
 $ kubectl get nodes
 $ kubectl get pods --all-namespaces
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,7 +54,7 @@ To verify everything worked you can run:
 $ export KUBECONFIG=/var/lib/kube-spawn/default/kubeconfig
 $ kubectl get nodes
 $ kubectl get pods --all-namespaces
-$ kubectl create -f 'https://github.com/kubernetes/kubernetes/blob/master/examples/guestbook/all-in-one/frontend.yaml'
+$ kubectl create -f 'https://raw.githubusercontent.com/kubernetes/kubernetes/master/examples/guestbook/all-in-one/frontend.yaml'
```
alban: I like to suggest `kubectl apply` instead of `kubectl create` so it will create or update as appropriate :)

PR #160, opened 2017-10-10
pkg/nspawntool: mount multiple directories for rktlet

dongsupark:
As rktlet needs to store data into both `/var/lib/rkt` and `/var/lib/rktlet`, we need to mount both directories into nspawn containers, to avoid potential issues of `no free space` inside containers.


dongsupark:
This issue was actually a side effect of a branch of rktlet that stores data under a wrong directory.  So this PR is not necessary. Closing.

PR #161, opened 2017-10-12
Refactor + CLI restructure

robertgzr:
The README is up to date (at least the quickstart section) so that can be used for testing as the way we interact with kube-spawn has changed quite a lot:

- `kube-spawn create`: used to generate all of the scripts and configs we need in the machines.
Everything is put into `/var/lib/kube-spawn/default` (tweakable with the `--dir` and `--cluster-name` flags).
It also outputs a `kspawn.toml` file into the root of that directory that contains the parameters used by kube-spawn (node number, base image, k8s version, etc.)
Most important things can be changed on create using flags and environment variables.

- `kube-spawn start`: takes the generated files and starts the machines and runs `kubeadm` on them. After this you should have a working cluster

- `kube-spawn stop`: brings down everything (does not retain the cluster as we delete the base images for the nodes)

- `kube-spawn destroy`: deletes `/var/lib/kube-spawn/default`


robertgzr:
CI fails because I don't have my [fork of `go-toml`](https://github.com/robertgzr/go-toml) checked into vendor/

pkg/config/types.go
```diff
@@ -0,0 +1,98 @@
...
+	// DevCluster indicates if we should run
+	// from a local kubernetes build
+	DevCluster bool `toml:"dev" mapstructure:"dev"`
+
+	RuntimeConfiguration RuntimeConfiguration `toml:"runtime-config,omitempty" mapstructure:"runtime-config"`
+
+	// Files to be copied from host to overlay of all machines
+	// map: dst = src
+	// not saved into kspawn.toml
+	Copymap map[string]string `toml:"-"`
```
dongsupark: Is there any reason to do it like `dst = src`? Because to me it looks intuitive to have the other way around, `src = dst`.

robertgzr: Yeah I thought about both ways... I ended up going with this because it feels like variable assignment to me. 
But I am open to changing it if that makes no sense.

dongsupark: @robertgzr Ok, if you say so, I don't have strong opinion.

robertgzr:
I tried splitting up the big commit into logical chunks, commited the vendor fixes, added a quick little list subcommand and fixed the panics related to running with `rkt` as container runtime

robertgzr:
Things that remain todo:
- [ ] clean up cni config (probably in a later PR)
- [ ] terminal output broken after `kube-spawn start`

robertgzr:
Rebased and replaced the delay in the stop commad with a retry loop

pkg/config/defaults.go
```diff
@@ -0,0 +1,257 @@
...
+	DefaultKubernetesVersion = "v1.7.5"
+	DefaultBaseImage         = "coreos"
+
+	DefaultDockerRuntimeEndpoint = "" // "unix:///var/run/docker.sock"
+	DefaultRktRuntimeEndpoint    = "unix:///var/run/rktlet.sock"
+	DefaultCrioRuntimeEndpoint   = "unix:///var/run/crio.sock"
+	DefaultRuntimeTimeout        = "15m"
+	DefaultRktStage1ImagePath    = "/usr/lib/rkt/stage1-images/stage1-coreos.aci"
+
+	CacheDir = ".cache"
```
schu: I don't think the directory should be hidden, as it leaves in `/var/lib/kube-spawn`.

pkg/config/defaults.go
```diff
@@ -0,0 +1,257 @@
...
+
+func SetDefaults_RuntimeConfiguration(cfg *ClusterConfiguration) error {
+	if cfg.RuntimeConfiguration.Timeout == "" {
+		cfg.RuntimeConfiguration.Timeout = DefaultRuntimeTimeout
+	}
+
+	// NOTE: K8s 1.8 or newer fails to run by default when swap is enabled.
+	// So we should disable the feature with an option "--fail-swap-on=false".
+	if !cfg.DevCluster && utils.CheckVersionConstraint(cfg.KubernetesVersion, "<1.8.0") {
+		cfg.RuntimeConfiguration.FailSwapOn = true
```
schu: I have asked this before: can we not disable swap for the machines instead?

dongsupark: As for the swap, I think it's hard to expect every user to disable swap. It might be easier to do that if we are only dealing with servers, or if most Linux distros disable swap out of the box. Sadly both are not the case.
So it's kind of nasty, but I would prefer to get it handled in this way on the k8s side.

cmd/kube-spawn/create.go
```diff
@@ -0,0 +1,165 @@
...
+	// and call from the if uninitialized {} block below
+	//
+	createCmd.Flags().StringP("container-runtime", "r", "", "runtime to use for the spawned cluster (docker or rkt)")
+	createCmd.Flags().String("kubernetes-version", "", `version kubernetes used to initialize the cluster. Irrelevant if used wih --dev. Only accepts semantic version strings like "v1.7.5"`)
+	createCmd.Flags().Bool("dev", false, "create a cluster from a local build of Kubernetes")
+	createCmd.Flags().IntP("nodes", "n", 0, "number of nodes to spawn")
+	createCmd.Flags().StringP("image", "i", "", "base image for nodes")
+	viper.BindPFlags(createCmd.Flags())
+
+	viper.BindEnv("runtime-config.rkt.rkt-bin", "KUBE_SPAWN_RKT_BIN")
```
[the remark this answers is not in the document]
robertgzr: maybe this is just me but I don't like the idea of having cli flags for all of these binaries. It's more or less necessary for the stage1 image but if you have installed rkt with your distros pkg manager and/or have `$GOPATH/bin` in your `PATH` you won't need to configure it anyway...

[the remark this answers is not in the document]
robertgzr: Also we have the problem that not all of the options in the config file are possible to be set from env/cli (or make sense to be) - like the path mapping for example.

This might be solvable by having two structures, one that viper binds to and the other we handle ourselves.

cmd/kube-spawn/start.go
```diff
@@ -0,0 +1,161 @@
...
+	if config.RunningMachines(cfg) == cfg.Nodes {
+		log.Print("cluster is up")
+		return
+	}
+
+	log.Printf("using %q base image from /var/lib/machines", cfg.Image)
+	log.Printf("spawning cluster %q (%d machines)", cfg.Name, cfg.Nodes)
+
+	// TODO: find a place for these calls
+	tmp(cfg)
```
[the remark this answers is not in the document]
robertgzr: will fix this

cmd/kube-spawn/stop.go
```diff
@@ -18,88 +18,117 @@ package main
...
-		cmd := exec.Cmd{
-			Path:   cmdPath,
-			Args:   args,
-			Env:    os.Environ(),
-			Stdout: os.Stdout,
-			Stderr: os.Stderr,
+func removeImage(machineName string) error {
+	done := false
+	retries := 0
+	for !done {
```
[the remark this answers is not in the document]
robertgzr: Looks like this also applies to https://github.com/kinvolk/kube-spawn/blob/master/pkg/nspawntool/run.go#L106

pkg/config/defaults.go
robertgzr: I wanted some kind of prefix as this makes distinguishing and [listing](https://github.com/kinvolk/kube-spawn/pull/161/commits/5b437dab9961f6c734701fd05c521aeecc0860a4#diff-04a5247871266df765044434b04e51acR58) cluster envs easier.

pkg/config/defaults.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	cniPath := os.Getenv("CNI_PATH")
+	if cniPath == "" {
+		return errors.New("CNI_PATH was not set")
+	}
+	pms := []Pathmap{
+		{Dst: "/usr/bin/rkt", Src: cfg.RuntimeConfiguration.Rkt.RktBin},
+		{Dst: "/usr/bin/rktlet", Src: cfg.RuntimeConfiguration.Rkt.RktletBin},
+		{Dst: "/usr/bin/stage1-coreos.aci", Src: cfg.RuntimeConfiguration.Rkt.Stage1Image},
```
iaguis: can we change this so it works if I use a non-coreos stage1? (see https://github.com/kinvolk/kube-spawn/pull/162)

robertgzr: forgot about this, thanks

robertgzr:
Ready for merge 🎉

pkg/config/defaults.go
schu: The better way would be then to have an `environments` subfolder. Can be changed in the next version.

cmd/kube-spawn/create.go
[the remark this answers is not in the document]
schu: Yeah, probably user and cluster configuration should be split.

blixtra:
I notice the `up` command was removed. Let's please add that back. We should have a one-command way to get the cluster from 0 to up.

We can add that back in another PR.

pkg/config/defaults.go
schu: I was thinking of using systemd properties (`systemd-nspawn --property ...`) to disable it for the containers, not for the host system :) On a cursory look, I haven't found a way to do this from the CLI though. So I'm not sure it's possible. IIRC LXD does provide an option for that.

Issue #154, opened 2017-09-29
check for versions of downloaded k8s binaries


dongsupark:
We should check for versions of downloaded k8s binaries. Otherwise we could have a lot version conflicts between downloaded ones and expected ones.

dongsupark:
Resolved via https://github.com/kinvolk/kube-spawn/pull/161

PR #164, opened 2017-10-16
bootstrap: check for versions of pre-existing binaries

dongsupark:
We need to check for versions of pre-existing binaries `/var/lib/kube-spawn/k8s/kube{ctl,let,adm}` to be able to avoid version mismatches between different k8s versions.

Versions are obtained by running commands like below:                                          

```                                                                                            
$ kubectl version --client=true --short=true                                                   
$ kubelet --version                                                                            
$ kubeadm version --output=short                                                               
```                                                                                            

Also add helpers checking for semantic versions like `IsSemVer()` or `IsSemVerOrNewer()`, and do some refactoring.

Fixes https://github.com/kinvolk/kube-spawn/issues/154

### DIFF
--- a/pkg/utils/semver.go
+++ b/pkg/utils/semver.go
@@ -1,0 +1,57 @@
+/*
+Copyright 2017 Kinvolk GmbH
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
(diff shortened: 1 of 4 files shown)


pkg/utils/semver.go
```diff
@@ -0,0 +1,51 @@
...
+// IsSemVer returns true if a semantic version string keyVersion is the same as
+// conVersion. For example, if both keyVersion and conVersion are 1.8.1,
+// it returns true.
+func IsSemVer(keyVersion, conVersion string) bool {
+	v, err := semver.NewVersion(keyVersion)
+	if err != nil {
+		return false
+	}
+
+	c, _ := semver.NewConstraint(fmt.Sprintf("=%s", conVersion))
```
nhlfr: Please check for error here

dongsupark: @nhlfr Added error checks.
BTW let's not merge it for now. Let's wait a while, and see how the PR https://github.com/kinvolk/kube-spawn/pull/161 goes on, because that PR was supposed to solve the same issue.

dongsupark:
Since https://github.com/kinvolk/kube-spawn/pull/161 was merged, this PR is not necessary any more.

PR #174, opened 2017-10-26
Reintroduce up command and improve cli formatting

robertgzr:


robertgzr:
@dongsupark but do we want to allow that?

I thought we wanted `up` to be the easy way to get everything running. 
For example we can't allow the `--image` option anyway, because we already provide the container linux base image.

If we want to define the flags from `create` as well I would like to allow all of them and let the user handle the image.

dongsupark:
Well, I'm still not completely sure if it's good enough to go without any options for `up`. But anyway we need to have the `up` command in any shape.
So I'll merge it for now.

Issue #11, opened 2017-04-10
clean up after cni


robertgzr:
- clean up iptables nat cni chains
```
sudo iptables -t nat -D POSTROUTING <line number>
sudo iptables -t nat -F <chain id>
sudo iptables -t nat -X <chain id>
```
- clean up cni ips
```
rm /var/lib/cni/networks/mynet/*
```

blixtra:
What's the status of this, @robertgzr?

robertgzr:
CNI should still create files inside `/var/lib/cni/networks/kube-spawn-net/`
Not sure about iptables

Maybe this can be part of the `destroy` command #38

PR #190, opened 2017-11-02
cmd/kube-spawn: remove cni config when destroying a cluster

dongsupark:
Let's make `./kube-spawn destroy` removes every CNI configuration file under `/var/lib/cni/networks/kube-spawn-net` as well as `/etc/cni/net.d/10-kube-spawn-net.conf`.

Fixes https://github.com/kinvolk/kube-spawn/issues/11


dongsupark:
@robertgzr Right. I just removed the option.

Issue #179, opened 2017-11-01
Split user and cluster configuration


schu:
Currently, kube-spawn uses a single configuration toml file for both user configuration (e.g. the k8s version to use) and node/cluster configuration (e.g. necessary bindmounts). This should be split.

robertgzr:
This issue corresponds to the conversation we had [here](https://github.com/kinvolk/kube-spawn/pull/161#discussion_r146816040), right?

> Also we have the problem that not all of the options in the config file are possible to be set from env/cli (or make sense to be) - like the path mapping for example.
> This might be solvable by having two structures, one that viper binds to and the other we handle ourselves.

The initial problem was that we can't set path mappings via flags so the user has to edit kspawn.toml to mount additional files.

Why do you want to split it?
I think it's fine to expose those things, the problem is more how we handle the options from config/flags/evs internally.

schu:
> This issue corresponds to https://github.com/kinvolk/kube-spawn/pull/161#discussion_r146816040

Yes.

> Why do you want to split it?

I would like to put configuration (e.g. k8s version to use, rktlet binary paths, ...) into configuration files and use those instead of long commands. Users then could also share kube-spawn config files where sensible.

schu:
This is done as part of https://github.com/kinvolk/kube-spawn/pull/265

PR #192, opened 2017-11-02
cmd/kube-spawn: add set-env command

robertgzr:
Allows to change the default cluster environment kube-spawn uses
when run without `--cluster-name`.

### DIFF
--- a/pkg/config/utils.go
+++ b/pkg/config/utils.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path"
 
@@ -32,6 +33,7 @@ const (
 	Filename = "kspawn.toml"
 
 	machineNameTemplate = "kubespawn%d"
+	envFilename         = "current-env"
 )
 
 func MachineName(no int) string {
@@ -51,10 +53,14 @@ func RunningMachines(cfg *ClusterConfiguration) int {
 }
 
 func LoadConfig() (*ClusterConfiguration, error) {
-	cfgFile := path.Join(viper.GetString("dir"), viper.GetString("cluster-name"), Filename)
+	env, err := GetCurrentEnv()
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to read current env")
+	}
+
+	cfgFile := path.Join(viper.GetString("dir"), env, Filename)
 	viper.SetConfigFile(cfgFile)
 
-	var err error
 	var cfg = &ClusterConfiguration{}
 	err = viper.ReadInConfig()
 	if err := viper.Unmarshal(cfg); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/config/utils.go
```diff
@@ -80,3 +86,21 @@ func WriteConfigToFile(cfg *ClusterConfiguration) error {
...
+func SetCurrentEnv(name string) error {
+	envFile := path.Join(viper.GetString("dir"), envFilename)
+	return fs.CreateBytes(envFile, []byte(name))
+}
+
+func GetCurrentEnv() (string, error) {
+	envFile := path.Join(viper.GetString("dir"), envFilename)
+	b, err := ioutil.ReadFile(envFile)
+	if IsNotFound(err) {
+		cName := viper.GetString("cluster-name")
```
schu: I think it should be possible to still explicitly set the env (`-c`), even when a default env is set.

schu: Alternatively, user configuration files could be implemented (see also #179). Then viper/cobra would take care of precedence automatically.

robertgzr: You can still set the environment using `--cluster-name=`. This is just implementing another way of setting it.
We should rename the `cluster-name` flag to `env` to get rid of the confusion.

> Alternatively, user configuration files could be implemented

The `env` option is not environment specific? In my opinion configuration files should be per-environment. This is why I implemented this as just a normal textfile...

schu: > You can still set the environment using `--cluster-name=`

`cluster-name` is only considered if the `current-env` file doesn't exist and it is created immediately if it can't be found.

Starting from an empty `/var/lib/kube-spawn`, you can see the problem with e.g.

```
sudo -E ./kube-spawn create -n 2
...
sudo -E ./kube-spawn create -n 3 -c foo
...
sudo -E ./kube-spawn start -c foo
...
spawning cluster "foo" (2 machines)
waiting for machine "kubespawn1" to start up
waiting for machine "kubespawn0" to start up
...
```

That's wrong and 3 nodes should have been spawned, right?

When looking at `destroy`, it looks like this:

```
sudo ./kube-spawn destroy -c foo
using config from /var/lib/kube-spawn/foo
destroying cluster "foo"
force stopping 2 machines
"kubespawn0" stopped
"kubespawn1" stopped
"foo" destroyed
```

Here it partly works: first `default/kspawn.toml` is loaded, but since this is done with viper, `cluster-name` as specified on the CLI is prioritized and the right directory is deleted from disk  (while it stops the wrong containers). The log is confusing since `cfg.Name` is used for that.

https://github.com/kinvolk/kube-spawn/blob/082cb673c4564ea57dc2d5db308a22123561e6ae/pkg/config/utils.go#L62-L65

I would do two things:

* IMHO user configuration should be configurable via cmdline, configuration files and environment variables. For this vanilla viper + cobra could be used, which gives you auto-loading, precedence handling, etc. out of the box. (The fact that viper seems to support only a single configuration file might requires restructuring the current config model a bit, to enable both global (e.g. default env) and per cluster config (e.g. container runtime).)
* Internal configuration can continue to live in `/var/lib/kube-spawn/<env>/kspawn.toml` but should not be loaded with viper (to avoid mixing user and internal configuration), i.e. avoid `viper.Unmarshal(cfg)` and instead populate `ClusterConfiguration` explicitly.

schu:
Doesn't work as is and should be approached differently, closing.

PR #193, opened 2017-11-03
pkg/script: change Docker config to use overlay2

jonboulle:
`overlay` is the old/deprecated overlayfs driver and does not work out
on top of btrfs, leading kube-spawn to fail out of the box when it
initialises a cluster on top of a btrfs filesystem.


jonboulle:
bump? hopefully this is fairly uncontroversial..

PR #200, opened 2017-11-10
pkg: clean up unused functions

dongsupark:
Clean up unused functions except for `CheckValidFile` which will be probably used later again.

These are detected by `codecoroner funcs --ignore vendor ./...`.


dongsupark:
@schu All right. I removed CheckValidFile too.

schu:
Thanks!

Issue #203, opened 2017-11-11
Instructions in README seem to be wrong


schu:
Currently, `start` fails with

```
failed to start machine: error running cnispawn: exec: "cnispawn": executable file not found in $PATH
```

Missing a documentation update after https://github.com/kinvolk/kube-spawn/pull/199

PR #204, opened 2017-11-11
README: add an instruction to install binaries

dongsupark:
Since https://github.com/kinvolk/kube-spawn/pull/199, it is possible to install kube-spawn binaries under a system directory.

Fixes https://github.com/kinvolk/kube-spawn/issues/203

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,11 @@ $ go get -d github.com/kinvolk/kube-spawn
 $ cd $GOPATH/src/github.com/kinvolk/kube-spawn
 $ make all
 
+# (optional) Install binaries under a system directory.
+# The install prefix defaults to /usr, which you can override with an env
+# variable $PREFIX, like "make PREFIX=/usr/local install".
+$ sudo make install
+
 $ export CNI_PATH=$GOPATH/bin
 
 # This generated a default 3 nodes cluster configuration


alban:
Is there a way to make it work with an environment variable without requiring `sudo make install`?

dongsupark:
@alban You can specify an env variable `BINDIR` to a target directory for `make install`, e.g. 
```
$ BINDIR=$HOME/bin make install
$ export PATH=$HOME/bin:$PATH
```
Or without installation, just like this:
```
$ export PATH=$GOPATH/src/github.com/kinvolk/kube-spawn:$PATH
```

krnowak:
I actually did not intend to break kube-spawn's behaviour about binaries - `PATH` was meant to be an additional source of binaries, so to speak. The PR #207 fixes that, so it makes `make install` optional for developing purposes.

nhlfr:
@krnowak's PR is merged now

@dongsupark please mention that the step with binaries is optional

dongsupark:
@nhlfr Done. Updated the words.

README.md
```diff
@@ -36,6 +36,9 @@ $ go get -d github.com/kinvolk/kube-spawn
 $ cd $GOPATH/src/github.com/kinvolk/kube-spawn
 $ make all
 
+# (optional) Install binaries under a system directory
```
schu: Can you say where they will be installed and how it can be changed?

dongsupark: @schu Done. Added some comments.

dongsupark:
@alban When I initially created this PR, it was true that it's not optional. Meanwhile, https://github.com/kinvolk/kube-spawn/pull/207 was made and already merged. With that, `make install` became optional again. That's why I added `(optional)` in this PR.

Issue #175, opened 2017-10-26
Provide kubectl


zeenix:
kube-spawn makes it really easy to get and launch local k8s for newbies but users most likely also want to have a kubectl binary to interact with the cluster. Currently AFAIK you are expected to get that yourself and most people would just get it from their distro and that will likely not have all the feature/parameters that the k8s fetched by kube-spawn will have.

So it would be nice if kube-spawn could also get the kubectl for users or at least make it easy to install the latest one.

krnowak:
Isn't kubectl installed in `/var/lib/kube-spawn/k8s/kubectl`? You could just adjust your PATH to point there.

alban:
For me it is in:
```
/var/lib/kube-spawn/default/rootfs/usr/bin/kubectl
```
But the path depends on the cluster name...

I think it is fine if users has to get `kubectl` themselves, but the README has to be updated to explain how to do that.

Example:
```
curl -Lo /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
chmod +x /usr/local/bin/kubectl
```

PR #209, opened 2017-11-14
readme: add instruction to get kubectl

alban:
Fixes https://github.com/kinvolk/kube-spawn/issues/175

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,13 @@ $ kubectl get pods --all-namespaces
 $ kubectl create -f 'https://raw.githubusercontent.com/kubernetes/kubernetes/master/examples/guestbook/all-in-one/frontend.yaml'
 ```
 
+If you don't have `kubectl`, you can get it with:
+```
+KUBERNETES_VERSION=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)
+sudo curl -Lo /usr/local/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/${KUBERNETES_VERSION}/bin/linux/amd64/kubectl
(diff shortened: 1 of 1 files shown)


zeenix:
But with kube-spawn you already have the binary, just need to add the binary to the path so this would be redundant.

krnowak:
Is it important to have a version of kubectl that matches a version of kubernetes kube-spawn is using?

Maybe it would be better to tell the user where kube-spawn puts kubectl and other tools, so the user can decide what to do with it.

alban:
> Is it important to have a version of kubectl that matches a version of kubernetes kube-spawn is using?

It does not seem to be too important: I have kubectl 1.8.3 installed on my laptop and it works with Kubernetes 1.7. Maybe the other way around could cause problems? I don't know.

> Maybe it would be better to tell the user where kube-spawn puts kubectl and other tools, so the user can decide what to do with it.

Do you mean to tell the user via the README.md or via a `fmt.Printf` after a `kube-spawn start`?

robertgzr:
What about `kube-spawn` checking the installed version and if there is a mismatch/ or no `kubectl` installed, notifying the user?

krnowak:
Unless you put the message in a nice box and some terminalesque `<blink>` tag. ;)

dongsupark:
@robertgzr Actually I already made a PR to check the versions: https://github.com/kinvolk/kube-spawn/pull/164
Then afterwards I just throw it away, as we introduced restructuring of configs. Right now we don't need to check it any more, as different versions are stored in different directories.

So I agree with adding just a simple description in README. I don't care about the details, but the description should be there. Even if there's no problem with version mismatches, there might be problems in the future.

Issue #213, opened 2017-11-16
Starting kube-spawn fails if the images are already there


iaguis:
Trying to start kube-spawn when the (cloned) node images are present results on errors.

```
# ./kube-spawn start
using config from /var/lib/kube-spawn/default
using "coreos" base image from /var/lib/machines
spawning cluster "default" (3 machines)
Image resized.
Resize '/var/lib/machines' of 'max'
waiting for machine "kubespawndefault2" to start up
waiting for machine "kubespawndefault0" to start up
waiting for machine "kubespawndefault1" to start up
failed to start machine: error cloning image: "/bin/machinectl clone coreos kubespawndefault0" failed: Could not clone image: File exists
failed to start machine: error cloning image: "/bin/machinectl clone coreos kubespawndefault1" failed: Could not clone image: File exists
failed to start machine: error cloning image: "/bin/machinectl clone coreos kubespawndefault2" failed: Could not clone image: File exists
not all machines started
```

The current workaround is running

```
$ sudo machinectl remove kubespawndefault{0,1,2}
```

I think we should remove them ourselves, or reuse them.

PR #214, opened 2017-11-16
cmd/kube-spawn: retry to remove image until it succeeds

dongsupark:
Sometimes `kube-spawn stop` succeeded, but afterwards kube-spawn is not able to remove `kubespawndefault*` images, because of errors like `device or resource busy`. That makes the next round of starting clusters fail because of the remaining images.

This happens simply because it takes some time until nspawn containers can release every resource. So we need to retry to remove image until it succeeds, up to 15 * 0.5 sec = 7.5 sec.

Fixes https://github.com/kinvolk/kube-spawn/issues/213


dongsupark:
FYI, one travis failure was just a flaky one, which you can ignore.

PR #215, opened 2017-11-21
[RFC] list: indicate when environments are invalid

alban:
Before kube-spawn supported multiple cluster environments,
/var/lib/kube-spawn used to contain directory such as "etc", "k8s" etc.
When updating to a newer kube-spawn, this is mixed with the different
environments. This makes the "kube-spawn list" command confused.

This patch makes it more obvious that "k8s" is not a real cluster name:
```
$ sudo -E kube-spawn list
ENV NAME            LAST MODIFIED
default             Nov 16 17:43:05
etc [INVALID]       Oct  9 16:39:36
extras [INVALID]    Sep 22 15:03:21
k8s [INVALID]       Oct  9 16:39:36
scripts [INVALID]   Sep 22 16:31:10

5 environment(s) found
```

To achieve this, the command list actually parses the kspawn.toml
configuration file and prints "[INVALID]" when the parsing fails. The
LoadConfig() function in 'pkg/config/utils.go' was changed to accept the
cluster name as parameter instead of using the viper flag. I think it's
better if code in 'pkg/' does not use viper in this way because
libraries should not have a clear API without hidden side-channels.
Besides, it will help when implementing a kube-spawn daemon with a gRPC
interface.

### DIFF
--- a/cmd/kube-spawn/create.go
+++ b/cmd/kube-spawn/create.go
@@ -82,7 +82,7 @@ func runCreate(cmd *cobra.Command, args []string) {
 }
 
 func doCreate() {
-	cfg, err := config.LoadConfig()
+	cfg, err := config.LoadConfig(viper.GetString("cluster-name"))
 	if err != nil {
 		// ignore if config not found
 		// it means we started from scratch and need to generate one
--- a/cmd/kube-spawn/kube-spawn.go
+++ b/cmd/kube-spawn/kube-spawn.go
@@ -68,7 +68,7 @@ func main() {
 }
 
 func loadConfig() *config.ClusterConfiguration {
-	cfg, err := config.LoadConfig()
+	cfg, err := config.LoadConfig(viper.GetString("cluster-name"))
 	if err != nil {
 		log.Fatal(errors.Wrap(err, "error loading config"))
 	}
--- a/cmd/kube-spawn/list.go
+++ b/cmd/kube-spawn/list.go
@@ -26,9 +26,11 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+
+	"github.com/kinvolk/kube-spawn/pkg/config"
 )
 
-const tableFmt = "%-10s  %s"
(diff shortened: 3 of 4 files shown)


schu:
I didn't immediately understand what the pull request does: it adds logic to detect a scenario where a user runs a current version of kube-spawn with an old `/var/lib/kube-spawn` directory structure. Today, kube-spawn supports multiple environments and maintains a sub directory for each of those in `/var/lib/kube-spawn`. In earlier versions, only one environment was supported and its data stored directly in `/var/lib/kube-spawn` (hence the various "invalid" directories in the example above).

I think we shouldn't apply this patch.

First, it doesn't help the user. Even when they have done everything by the documentation until now, they suddenly see themselves confronted with the message. So they'll have to open an issue or dig into the code until they find out what's going on or not think about it and wipe `/var/lib/kube-spawn`. (Side note: it says `5 environment(s) found` while at the same time marking 4 as invalid, i.e. no environments?)

Second, it treats a symptom and not the cause.

Instead, users should learn from the release notes about breaking changes and necessary steps to upgrade. To help users with recognizing breaking changes with regard to kube-spawn's internal data in `/var/lib/kube-spawn`, we could maintain a way that allows kube-spawn to identify the "version" of the data and act accordingly. IMHO, a better, practical user experience would be something like this:

```
$ sudo -E ./kube-spawn ....
Warning: kube-spawn has detected outdated internal data in /var/lib/kube-spawn which requires an upgrade
Please see the release notes at https://github.com/... to learn how to do that
Aborting
```

One way could be to add a file `/var/lib/kube-spawn/meta.json` where such a version is maintained.

dongsupark:
It's actually quite a corner case, so I don't care much whether it's applied or not.
Ideally @schu's idea is a good approach, but @alban's approach is also not bad from the user's perspective. I don't think that this patch is something that we should not apply.

alban:
I updated the release message on https://github.com/kinvolk/kube-spawn/releases/tag/v0.2.0

schu:
Outdated, closing.

PR #218, opened 2017-11-23
Remove leftovers after docker removal

nhlfr:
kube-spawn vagrant environment does not use
docker anymore for builds. This commit removes
all the leftovers after docker in vagrant provision
script.


alban:
But to build Kubernetes locally (and use it with `--dev`), we need docker to build the hyperkube image, right?

dongsupark:
@alban True.
However, in order to use Vagrant VMs for smoke tests on CI services, we should remove docker & k8s installation parts from the vagrant scripts. Otherwise each build would cause a lot of dependencies to be fetched. That's why I wanted to deliberately remove the installations.
In the future we could either create a separate setup script for `--dev` or add relevant documentation.

PR #219, opened 2017-11-23
doc/devel/release.md: updates

alban:
Add smoke test, check version.

-----

~I would like to merge this only after v0.2.0, so if I notice anything else during the release, I can add it to this PR.~

### DIFF
--- a/doc/devel/release.md
+++ b/doc/devel/release.md
@@ -23,18 +23,20 @@ The following example assumes we're going from version 0.1.1 (`v0.1.1`) to 0.2.0
 
 Let's get started:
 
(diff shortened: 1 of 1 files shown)


schu:
Other projects often have a release branch and code freeze period (i.e. time to actually use and test what will be the next release) before the release is made. I think it would be good to do that as well.

alban:
@schu rkt does that but in kube-spawn's case, there would be no commit on that release branch: the version in `kube-spawn --version` is compiled in from
```
VERSION=$(shell git describe --tags --always --dirty)
```
and
```
-ldflags "-X main.version=$(VERSION)"
```

So I removed it when copying the release guide from rkt.

dongsupark:
This PR updates only a document, so it does not conflict with the refactoring PR https://github.com/kinvolk/kube-spawn/pull/265 .
So merge.

PR #222, opened 2017-11-25
kubeadm join: fix TLS on Kubernetes >=1.9

alban:
kubeadm got a new option in version 1.8: `--discovery-token-unsafe-skip-ca-verification`
See: https://github.com/kubernetes/kubernetes/pull/49520

Since version 1.9, it became mandatory to either use it to skip verification, or to use `--discovery-token-ca-cert-hash=...`
See: https://github.com/kubernetes/kubernetes/pull/55468

Since kube-spawn is a developer tool used on one physical machine, use the former.

### DIFF
--- a/pkg/nspawntool/kubeadm.go
+++ b/pkg/nspawntool/kubeadm.go
@@ -7,6 +7,7 @@ import (
 	"github.com/kinvolk/kube-spawn/pkg/bootstrap"
 	"github.com/kinvolk/kube-spawn/pkg/config"
 	"github.com/kinvolk/kube-spawn/pkg/machinetool"
+	"github.com/kinvolk/kube-spawn/pkg/utils"
 	"github.com/kinvolk/kube-spawn/pkg/utils/fs"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


blixtra:
Can you provide a description?

alban:
@blixtra updated.

pkg/nspawntool/kubeadm.go
```diff
@@ -88,7 +89,18 @@ func JoinNode(cfg *config.ClusterConfiguration, mNo int) error {
...
-		"--token", cfg.Token,
-		cfg.Machines[0].IP + ":6443"}...)
+		"--token", cfg.Token}...)
+
+	// --discovery-token-unsafe-skip-ca-verification appeared in Kubernetes 1.8
+	// See: https://github.com/kubernetes/kubernetes/pull/49520
+	// It is mandatory since Kubernetes 1.9
+	// See: https://github.com/kubernetes/kubernetes/pull/55468
+	// Test is !<1.8 instead of >=1.8 in order to handle non-semver version 'latest'
+	if !utils.CheckVersionConstraint(cfg.KubernetesVersion, "<1.8") {
```
dongsupark: FYI, similar version checks are done in `pkg/config/defaults.go`, either `if cfg.DevCluster || utils.CheckVersionConstraint(cfg.KubernetesVersion, ">=1.8.0")` or `if !cfg.DevCluster && utils.CheckVersionConstraint(cfg.KubernetesVersion, "<1.8.0")`. In this file the check is now used a little differently. Basically this pattern is kind of error-prone, so I think we should be somehow consistent. We could make a helper to simplify it in another PR.

alban:
Thanks for the review & test!

PR #224, opened 2017-11-27
vagrant: Expose apiserver outside the VM

### LINKED ISSUE #217: [vagrant] Provide port-forwarding from the k8s api to the host
It would be nice to have the kube-spawn cluster in the VM accessible from the host for that we need to be able to access the k8s API via a port on the guest that we can forward to the VM host.

Possible ways to achieve this:
* run `socat`
* write `iptables` rules

nhlfr:
Introduce the KUBESPAWN_REDIRECT_TRAFFIC environment
variable to the build.sh script which redirects the
traffic incoming to the 6443 port in Vagrant VM to
the nspawn container with kube-apiserver.

Fixes #217

### DIFF
--- a/vagrant-all.sh
+++ b/vagrant-all.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -eo pipefail
+
+export KUBESPAWN_AUTOBUILD="true"
+export KUBESPAWN_DISTRO=${KUBESPAWN_DISTRO:-fedora}
+export KUBESPAWN_REDIRECT_TRAFFIC="true"
+
+KUBESPAWN_PROVIDER=${KUBESPAWN_PROVIDER:-virtualbox}
+
+vagrant up $KUBESPAWN_DISTRO --provider=$KUBESPAWN_PROVIDER
+
+./vagrant-fetch-kubeconfig.sh
+
+export KUBECONFIG=$(pwd)/kubeconfig
--- a/vagrant-fetch-kubeconfig.sh
+++ b/vagrant-fetch-kubeconfig.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+set -eo pipefail
+
+KUBESPAWN_DISTRO=${KUBESPAWN_DISTRO:-fedora}
+
+# The following command sometimes exists with status 1 when using vagrant-libvirt,
+# despite the fact that ssh config was generated successfully.
+vagrant ssh-config $KUBESPAWN_DISTRO > $(pwd)/.ssh_config || true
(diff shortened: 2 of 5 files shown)


robertgzr:
Have not tested it yet...

The scripts should probably go into the `/scripts` folder?

nhlfr:
Well, `/scripts` folder has only scripts are **runned by** Vagrant (during provisioning), so I thought that scripts which **run Vagrant** should be somewhere else. But I don't mind if you are fine with putting everything there.

vagrant-all.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -eo pipefail
+
+export KUBESPAWN_AUTOBUILD="true"
+export KUBESPAWN_REDIRECT_TRAFFIC="true"
+vagrant up
```
dongsupark: I would assume that it runs with fedora, as `vagrant up` without machine name defaults to fedora.

vagrant-fetch-kubeconfig.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -eo pipefail
+
+# The following command sometimes exists with status 1 when using vagrant-libvirt,
+# despite the fact that ssh config was generated successfully.
+vagrant ssh-config > $(pwd)/.ssh_config || true
```
dongsupark: You need to run `vagrant ssh-config fedora`. Otherwise it will just fail.

vagrant-all.sh
```diff
@@ -0,0 +1,11 @@
...
+
```
dongsupark: Are we missing something here for actually launching kube-spawn clusters? Because `vagrant up` launches a VM, but it does not run a kube-spawn cluster inside the VM. If `KUBESPAWN_AUTOBUILD` is true, the script should run something like `~/build.sh` or `kube-spawn create ; kube-spawn start` AFAIK.

nhlfr: It actually runs a kube-spawn cluster inside the VM. Please take a look again at Vagrantfile and setup-env script - they both consume those variables. When `KUBESPAWN_AUTOBUILD` is `true`, then `build.sh` script is running in the provisioning phase.

vagrant-fetch-kubeconfig.sh
nhlfr: Good point. This is connected with multi-distro thing you mentioned in the first comment. Will fix it soon.

vagrant-all.sh
dongsupark: Aha, I see.
Hmm, I'm not sure I like this autobuild logic being inside Vagrantfile, 3 times in a row. But if it works, I'm not against it for now.

nhlfr:
@dongsupark PTAL

PR #232, opened 2017-11-30
 doc/troubleshooting: how to trouble shoot inotify problems 

iaguis:
This happens when trying to run many nodes.

Also, add a TOC so users have an overview of the problems addressed.


iaguis:
Good point, I'll try to rephrase

iaguis:
Updated.

Issue #230, opened 2017-11-30
KUBECONFIG not set when running with fish shell


blixtra:
`kubectl get nodes` does not work without manually setting the KUBECONFIG variable when using the fish shell.

Had to run `export KUBECONFIG=/var/lib/kube-spawn/default/kubeconfig` to get it working.

robertgzr:
I don't think we set `KUBECONFIG` at all. 

The README says:
> To verify everything worked you can run:
> ```
> $ export KUBECONFIG=/var/lib/kube-spawn/default/kubeconfig
> $ kubectl get nodes
> ```

blixtra:
In the [Quick Start](https://github.com/kinvolk/kube-spawn/blob/master/README.md#quickstart) section, it does not say that.

We need to fix that.

robertgzr:
Should setting it be part of `up`?

I don't think `kube-spawn start` should set it because we can run multiple clusters and it would always overwrite it for the most recent one.

alban:
This is a matter of documentation because it's not possible for `kube-spawn` to set the environment variable of a parent process.

blixtra:
There used to be a hint telling the user what to do. That has disappeared. I'd suggest we add it back. Otherwise I assume it's done for me.

dongsupark:
@blixtra True. The hint about KUBECONFIG at the end of start has somehow disappeared.
I think that the message is indeed helpful. Let's bring it back.

alban:
It was deleted in https://github.com/kinvolk/kube-spawn/pull/161/files#diff-df46f5e52ac4d342643dc4c045faee8aL131:
```
-       log.Println("Note: For kubectl to work, please set $KUBECONFIG to " + utils.GetValidKubeConfig())
```

When adding it back, please give a line that can be copy-pasted in the shell, such as:
```
log.Println("Note: For kubectl to work, please set $KUBECONFIG:")
log.Printf("export KUBECONFIG=%s\n", utils.GetValidKubeConfig())
```
It was very annoying to type this manually.

alban:
#233 does not seem to work properly (at least on @blixtra's computer)

dongsupark:
@alban Right.
I saw too that the hint didn't get printed yesterday on @blixtra's computer, while it worked fine with me.

[...]

PR #233, opened 2017-11-30
cmd/kube-spawn: bring back help message for KUBECONFIG

dongsupark:
This help message for setting `$KUBECONFIG` at the end of `kube-spawn start` used to exist, but it was removed by https://github.com/kinvolk/kube-spawn/pull/161.
We should bring it back for better user experience.

Fixes https://github.com/kinvolk/kube-spawn/issues/230


alban:
It does not seem to work properly (at least on @blixtra's computer)

PR #237, opened 2017-12-01
pkg/utils: figure out kubeconfig path correctly with a cluster name

dongsupark:
Now that kube-spawn is able to deal with multiple clusters, path to kubeconfig, `$KUBECONFIG`, can also differ depending on the cluster name. So `utils.GetValidKubeConfig()` should take a parameter for the cluster name, to compose path to kubeconfig correctly. For example, if the cluster's name is `mycluster`, KUBECONFIG will be `/var/lib/kube-spawn/mycluster/kubeconfig`.

Fixes https://github.com/kinvolk/kube-spawn/issues/230

### DIFF
--- a/cmd/kube-spawn/start.go
+++ b/cmd/kube-spawn/start.go
@@ -121,7 +121,7 @@ func doStart(cfg *config.ClusterConfiguration, skipInit bool) {
 	}
 	log.Printf("cluster %q initialized", cfg.Name)
 	log.Println("Note: For kubectl to work, please set $KUBECONFIG:")
-	log.Printf("export KUBECONFIG=%s\n", utils.GetValidKubeConfig())
+	log.Printf("export KUBECONFIG=%s\n", utils.GetKubeconfigPath(cfg.KubeSpawnDir, cfg.Name))
 	saveConfig(cfg)
 }
 
--- a/pkg/utils/kubernetes.go
+++ b/pkg/utils/kubernetes.go
@@ -29,19 +29,14 @@ import (
 )
 
 const (
-	kubeSpawnDir string = "/var/lib/kube-spawn"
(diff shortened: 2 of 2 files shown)


schu:
Tested with `sudo -E ./kube-spawn create  -c c1 && sudo -E ./kube-spawn start  -c c1`, seems to work.

I didn't test with a custom data dir (`-d`) due to https://github.com/kinvolk/kube-spawn/issues/238 From a cursory look on the code, I assume it doesn't always work as expected, since data dir is not considered.?

Also, I wonder why https://github.com/kinvolk/kube-spawn/pull/233 reintroduced `CheckValidFile` (which is [buggy](https://github.com/kinvolk/kube-spawn/pull/200#issuecomment-343917333)). I also find the code rather difficult to read, for example due the subtle difference between `kcRelPath` and `ksRelPath`. The name `GetValidKubeConfig` suggests the return value is a validated kubeconfig object, where it actually is a string which could be empty (errors silently ignored) or a path to a file which could be a valid kubeconfig. I think a better function would be something like `GetKubeconfigPath(clusterName string) (string, error)`.

dongsupark:
Ok, the directory issue looks real. I need to fix it.

As for the `CheckValidFile()`, there's no special reason to have it again. It was just easy to bring back the old function. I admit that it's not the best code, but I don't have a better idea. If we would change this function, we should also update `CheckValidDir()` for consistency. That's way out of scope of this PR. Patches are welcome!

Yesterday I already thought about your comment that the function was buggy. So I tried to test it with different directory permissions, but couldn't find a particular issue. Maybe it's because we already need root permissions for running kube-spawn. Of course if there's still any potential bug, we should definitely fix it in the future.

As for other naming issues, I'll try to fix.

Issue #231, opened 2017-11-30
Running kube-spawn without sudo on Fedora causes authentication loop


blixtra:
I mistakenly ran `kube-spawn` without using `sudo` and was presented with a never ending authentication box. It required killing the terminal from another tty to get out of it.

This is a super really experience. We should check the `kube-spawn` is run with the proper permissions and refuse to run otherwise.

PR #248, opened 2017-12-07
cmd/kube-spawn: abort if it's non-root user

dongsupark:
Ordinary commands, create, destroy, restart, start, stop, and up, need to run with root privileges. So we need to exit if euid is not 0.

Though we need to exclude the list command, as non-root users are able to read profiles.

Fixes https://github.com/kinvolk/kube-spawn/issues/231

### DIFF
--- a/cmd/kube-spawn/create.go
+++ b/cmd/kube-spawn/create.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"golang.org/x/sys/unix"
 
 	"github.com/kinvolk/kube-spawn/pkg/bootstrap"
 	"github.com/kinvolk/kube-spawn/pkg/config"
@@ -79,6 +80,10 @@ func init() {
 }
 
 func runCreate(cmd *cobra.Command, args []string) {
+	if unix.Geteuid() != 0 {
+		log.Fatalf("non-root user cannot create clusters. abort.")
+	}
+
 	if len(args) > 0 {
 		log.Fatalf("too many arguments: %v", args)
 	}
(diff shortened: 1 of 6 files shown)


dongsupark:
As it's a relatively obvious bugfix. I'll just merge it.

schu:
An alternative would be to do this with a `PersistentPreRunE` function from the parent cmd like

```
diff --git a/cmd/kube-spawn/kube-spawn.go b/cmd/kube-spawn/kube-spawn.go
index 86e0572..3de017f 100644
--- a/cmd/kube-spawn/kube-spawn.go
+++ b/cmd/kube-spawn/kube-spawn.go
@@ -25,6 +25,7 @@ import (
        "github.com/pkg/errors"
        "github.com/spf13/cobra"
        "github.com/spf13/viper"
+       "golang.org/x/sys/unix"
 )
 
 var (
@@ -56,8 +57,26 @@ func init() {
 
        kubespawnCmd.Flags().BoolVarP(&printVersion, "version", "V", false, "Output version information")
 
-       kubespawnCmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {
+       kubespawnCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+               switch cmd.Use {
+               case "create":
+                       fallthrough
+               case "destroy":
+                       fallthrough
+               case "restart":
+                       fallthrough
+               case "start":
+                       fallthrough
+               case "stop":
+                       fallthrough
+               case "up":
+                       if unix.Geteuid() != 0 {
+                               cmd.SilenceUsage = true
+                               return fmt.Errorf("non-root user cannot %s clusters, aborting", cmd.Use)
+                       }
+               }
                log.SetFlags(0)
+               return nil
        }
 }
```

Issue #243, opened 2017-12-04
Improve k8s version checking patterns


dongsupark:
There are several Kubernetes version checks done in for example `pkg/config/defaults.go`, either `if cfg.DevCluster || utils.CheckVersionConstraint(cfg.KubernetesVersion, ">=1.8.0")` or `if !cfg.DevCluster && utils.CheckVersionConstraint(cfg.KubernetesVersion, "<1.8.0")`. In `pkg/nspawntool/kubeadm.go` the check is used a little differently like `if !utils.CheckVersionConstraint(cfg.KubernetesVersion, "<1.8")`.

Those patterns are basically error-prone, so I think we should make them somehow general and consistent. We could write a helper to simplify it.

See also https://github.com/kinvolk/kube-spawn/pull/222#discussion_r153077394

dongsupark:
Not valid at the moment. Closing.

PR #249, opened 2017-12-08
[do not merge] pkg/config: split out version checking function into IsVerConstraintOrDev

dongsupark:
To able to avoid error-prone version checking logic, we need to split it out into `IsVerConstraintOrDev()`. The function returns true if it's dev cluster, or if the kubernetes Version is "latest" (non-semver string), or the kubernetes version sastifies the given  constraint. That way we can simplify call sites as much as possible.

Also add a unit test to cover different cases of input parameters.

Fixes https://github.com/kinvolk/kube-spawn/issues/243

### DIFF
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -99,7 +99,7 @@ func SetDefaults_Kubernetes(cfg *ClusterConfiguration) error {
 		{Dst: "/usr/bin/socat", Src: path.Join(cacheDir, "socat")},
 	}
 
-	if cfg.DevCluster || utils.CheckVersionConstraint(cfg.KubernetesVersion, ">=1.8.0") {
+	if cfg.IsVerConstraintOrDev(">=1.8.0") {
 		cfg.TokenGroupsOption = "--groups=system:bootstrappers:kubeadm:default-node-token"
 	}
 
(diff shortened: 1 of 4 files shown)


dongsupark:
As https://github.com/kinvolk/kube-spawn/pull/259 depends on this PR, I'll merge this PR tomorrow, unless there's any objection.

schu:
Code like `if !cfg.IsVerConstraintOrDev(">=1.8.0") {` makes kube-spawn's code very difficult to read and understand. If I would have to guess what the functions does (when not allowed to look at the code), I would most certainly guess wrong, as the function name together with the argument doesn't tell the user what's happening ("Yes, `>=1.8.0` is a version constraint", "No clue what `Dev` is"), `. That's often an indicator for a design problem.

A "better" way (ignoring `cfg.Dev` for a moment) would be something like

```
if cfg.KubernetesVersion != "latest" && !doesSatisfyVersionConstraint(cfg.KubernetesVersion, ">=1.8.0") {
   // error out
}
```

Here ^ a new reader to the code would probably understand what it does from a glance w/o actually looking at the function.

I know the problem is inherent to kube-spawn's whole code base, so even if we would refactor this function, we still would have an overloaded `cfg.KubernetesVersion` field, the tangled `cfg` object and many other functions with non-self-explanatory names and implicit side effects on `cfg`.

Personally, I would not merge this and wait until we have refactored the code base (I'm taking a stab at this currently). But I wouldn't despair if you decide to merge it - as you want :)

dongsupark:
@schu Ok, then probably I'll need to keep this PR still open, and make https://github.com/kinvolk/kube-spawn/pull/259 independent of this PR. Feel free to open a new PR for replacing this PR, if you have a better idea.

TBH I don't think that in the future we are able to have enough time for refactoring the whole kube-spawn code base. So I'd like to keep individual refactoring items independent of each other.

dongsupark:
This PR will collide with the refactoring PR https://github.com/kinvolk/kube-spawn/pull/265. So let's not merge it for now.

dongsupark:
Will close this PR in favor of https://github.com/kinvolk/kube-spawn/pull/265

PR #250, opened 2017-12-08
cmd,pkg: upgrade the default Kubernetes version to v1.8.5

dongsupark:
Set the default Kubernetes version to v1.8.5, as it's k8s v1.8.x tree has been stable since months.


dongsupark:
As it's a trivial change, I'll merge it.
Users can still set the version back to lower versions, if they want to. So there's no need to worry about regression.

Issue #107, opened 2017-08-03
Verify downloaded images with gpg or sha512


dongsupark:
We need to verify downloaded images with either gpg or sha512.
See also https://github.com/kinvolk/kube-spawn/pull/106#issuecomment-319966741

PR #251, opened 2017-12-09
Check GnuPG signature of CoreOS image

dongsupark:
Verify GnuPG signature of downloaded CoreOS image. systemd-nspawn/machinectl doesn't have a functionality of checking downloaded images by GPG signatures. We need to download the image and validate it by ourselves and then import it by machinectl.

Also introduce a command line option `--image-gpg-verify` for kube-spawn create. This option is set to true by default. Any user who wants to disable gpg verification can set this option to false: `--image-gpg-verify=false`.

This PR is based on https://github.com/kinvolk/kube-spawn/pull/131 by @nhlfr. First I rebased it on top of master, which needed a bit of work, as the project structure has changed a lot recently. After that I decided to simplify the openpgp logic. Please let me know if anything is wrong. A couple of test results looked good to me.

Supersedes https://github.com/kinvolk/kube-spawn/pull/131
Fixes https://github.com/kinvolk/kube-spawn/issues/107


pkg/bootstrap/node.go
```diff
@@ -745,57 +753,179 @@ func checkCoreosVersion() error {
 	return nil
 }
 
-func pullRawCoreosImage() error {
-	var cmdPath string
-	var err error
+func ensureCoreosVersion() {
+	if err := checkCoreosVersion(); err != nil {
+		log.Println(err)
+		log.Fatalf("You will need to remove the image by 'sudo machinectl remove coreos' then the next run of kube-spawn will download version %s of coreos image automatically.", coreosStableVersion)
```
schu: Do we really need to stop here? Can we do the work for the user?

pkg/machinetool/machinetool.go
```diff
@@ -74,6 +74,11 @@ func Terminate(machine string) error {
 	return err
 }
 
+func ImportRaw(imagePath, imageName string) error {
+	_, err := machinectl(nil, nil, "--verify=no", "import-raw", imagePath, imageName)
```
schu: Why is machinectl's [`--verify`](https://www.freedesktop.org/software/systemd/man/machinectl.html) not an option for verification?

pkg/bootstrap/node.go
```diff
@@ -745,57 +753,179 @@ func checkCoreosVersion() error {
...
+func downloadSignature() error {
+	return downloadFile(signatureTmpFile, signatureUrl)
+}
+
+func verifyImage() error {
+	// TODO(nhlfr): Try to use or implement some library for managing PGP keys instead
+	// of executing gpg binary.
+	gpgCmdPath, err := exec.LookPath("gpg")
+	if err != nil {
+		gpgCmdPath = "/usr/bin/gpg"
```
schu: Some systems have both `gpg` and `gpg2`. Not sure picking `/usr/bin/gpg` is always right.

pkg/machinetool/machinetool.go
dongsupark: > Why is machinectl's --verify not an option for verification?

Because systemd only supports special name of GPG key files, which are not supported from CoreOS. AFAIK that was the whole point of writing this code in this place.

Quote from machinectl(1), in the section of pull-tar/pull-raw.

> Verification is done via SHA256SUMS and SHA256SUMS.gpg files that need to be made available on the same web server, under the same URL as the .tar file, but with the same web server, under the same URL as the .tar file, but with the last component (the filename) of the URL replaced. 

Maybe @nhlfr could give us more background.

dongsupark:
> But most importantly: like this, kube-spawn overwrites my $HOME/.gnupg with a new keyring, since it's run with sudo -E (i.e. $HOME points to /home/user).
> So since I haven't looked at the code carefully before testing, this just wiped out my user's GPG keyring and setup and I need to recover it from backup :(

Whoa, sorry about that. :(  We need to be able to avoid this case.

PR #256, opened 2018-01-10
Various improvements, updates and fixes

schu:
More to come..


alban:
Reviewed up to the `CreateDir` patch and it LGTM. I didn't review the last patch.

schu:
Ready for review / merge. Details can be found in the individual commit messages.

A quick recap of the main changes done:

* Remove `cni-noop` and `EnsureBridge`: it's not really clear why those were needed in the first place. kube-spawn uses a simple nspawn wrapper called `cnispawn` to setup containers which is responsible for creating the `cni0` bridge and should return an error if that doesn't work.
* `cnispawn` used to be a separate binary and is now a hidden subcommand (`cni-spawn`) to reduce the number of executables people have to install.
* Remove the awkward requirement for a `GOPATH` environment variable / `$GOPATH/src/kinvolk/kube-spawn/...` and simply load `kube-spawn-runc` from the current directory or, if not found there, from `$PATH`.
* Remove the requirement for a `CNI_PATH` environment variable and instead allow users to either a) put the CNI plugins into the default location `/opt/cni/bin` or b) use `kube-spawn create --cni-plugin-dir ...` to explicitly set a path for a cluster.

Please note that this is just a start to unblock coming work. In my opinion the code needs to be refactored entirely. I plan to gradually do that.

dongsupark:
In most cases, I agree with the motivations mentioned in this PR. I think that most of them are good catches. Thanks.

Some commits have independent things together, while other commits are split into somehow too small pieces. Though I don't care much about that. It would be really great if this PR could be split into multiple PRs, for example, one for obvious bugfixes, and the other for refactoring, etc. But I think it's also ok with the current form, as we are not tightly bound to certain release schedules. It's a perfect moment for refactoring. :-)

All I care is to make sure that the whole kube-spawn works without regression. I've just tested this PR with a simple setup, and it seems to work fine. Though it's not a complete test. I'll try to run it again later.

schu:
@dongsupark thanks for taking a look. Everything should work as before (TM) functionality wise.

@iaguis can you maybe smoke test with rktlet once?

I would love to merge later today or tomorrow.

iaguis:
> @iaguis can you maybe smoke test with rktlet once?

OK

iaguis:
No blockers from my side :)

Issue #212, opened 2017-11-16
'kube-spawn start' does not work as a systemd service


alban:
To start the nodes, `kube-spawn start` executes `systemd-nspawn` directly without `systemd-run`. So the `systemd-nspawn` processes stay in the same cgroup. When I run `systemd-run kube-spawn start` the cgroup could be for example `/system.slice/run-r6c31c44087e7466ba0a5174e650f4777.service`.

As soon as the command `kube-spawn start` finishes, systemd detects that the main pid of the service terminated so it kills all other processes of the cgroup, like `systemd-nspawn`. In effect, it kills the nodes.

Instead, `pkg/cnispawn/spawn.go:Spawn()` should start `systemd-nspawn` via `systemd-run`, so that the lifecycle of the nodes are not linked to the lifecycle of the `kube-spawn start` command.

This should work both when `kube-spawn start` is executed from the session in a terminal and as a systemd service.

PR #257, opened 2018-01-19
[WIP] cnispawn: run container in detached transient unit

schu:
With that, the container lifetime is not tied to the lifetime of the
slice in which kube-spawn is running.

This commit doesn't add code to shutdown and remove the network
namespace when the cluster is stopped. This is yet to be done.

### DIFF
--- a/pkg/cnispawn/spawn.go
+++ b/pkg/cnispawn/spawn.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cnispawn
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"runtime"
@@ -34,7 +35,6 @@ func Spawn(cniPluginDir string, nspawnArgs []string) error {
 	if err := cniNetns.Set(); err != nil {
 		return err
 	}
-	defer cniNetns.Close()
 
 	systemdNspawnPath := os.Getenv("SYSTEMD_NSPAWN_PATH")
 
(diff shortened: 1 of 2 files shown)


schu:
> This commit doesn't add code to shutdown and remove the network
namespace when the cluster is stopped. This is yet to be done.

I'm not sure yet how to do that best and maybe want to combine it with refactoring the code. Input welcome.

schu:
xref https://github.com/kinvolk/kube-spawn/issues/212

alban:
kube-spawn should check if the version of systemd-nspawn on the host is [v236](https://github.com/systemd/systemd/blob/v236/NEWS#L198-L200) before using `--network-namespace-path=`.

rkt has a [similar version test](https://github.com/rkt/rkt/blob/275da72499e6e3aa36f2ffe99cfa281179e61fbf/stage1/init/init.go#L415-L428).

Unfortunately, we need fallback code because v236 is not in every distro.

pkg/cnispawn/spawn.go
```diff
@@ -45,26 +45,31 @@ func Spawn(cniPluginDir string, nspawnArgs []string) error {
...
 
 	args = append(args, nspawnArgs...)
 
-	env := os.Environ()
-	env = append(env, "SYSTEMD_NSPAWN_MOUNT_RW=1")
-	env = append(env, "SYSTEMD_NSPAWN_API_VFS_WRITABLE=1")
-	env = append(env, "SYSTEMD_NSPAWN_USE_CGNS=0")
-
-	_, err = syscall.ForkExec(systemdNspawnPath, args, &syscall.ProcAttr{
+	_, err = syscall.ForkExec(systemdRunPath, args, &syscall.ProcAttr{
```
alban: We could use `systemd-run --service-type=notify` together with `systemd-nspawn --notify-ready=yes` right now. We could then debug more easily with `systemctl status run-xxxx.service`.

Possibly use `systemd-run --unit=` using a user-friendly generated name too. And `--slice=machine.slice` (I had a look at [/usr/lib/systemd/system/systemd-nspawn@.service](https://github.com/systemd/systemd/blob/v236/units/systemd-nspawn%40.service.in#L25)).

And lastly, but it might require more refactoring so maybe in a future step: instead of `syscall.ForkExec()`, use [os.exec](https://golang.org/pkg/os/exec/#Cmd) so you can catch the stdout/stderr and exit status, and print those in case of errors. We would need to start the 3 nodes in parallel (in case of a 3-node cluster) and I don't know if the current code does this in parallel (hence the need for refactoring).

dongsupark:
I've just realized that this PR would conflict with my PR https://github.com/kinvolk/kube-spawn/pull/328. I totally forgot.
I think we should make a new issue to address points that were mentioned here. Then after #328 was merged, we can revisit those issues.
Is that ok?

PR #261, opened 2018-01-25
vendor: clean up unnecessary vendor trees by using dep 0.4

dongsupark:
Clean up unused pakages, documents, tests, etc, by using `[prune]` directive introduced in dep 0.4. The essential part is the following configs in `Gopkg.toml`.

```
[prune]
  go-tests = true
  non-go = true
  unused-packages = true
```


schu:
Great!

Issue #255, opened 2017-12-27
Nodes not ready on vagrant box after halting and starting VM again


kosyfrances:
Running `kubectl get nodes` for the first time in a vagrant box works fine. Same with `kubectl get pods --all-namespaces`. 

The challenge comes in when I halt the VM and spin it up again, all the nodes are never ready. 
Here is the output of `kubectl get nodes`
```
NAME                STATUS     ROLES     AGE       VERSION
kubespawndefault0   NotReady   master    25m       v1.8.5
kubespawndefault1   NotReady   <none>    25m       v1.8.5
```
and that of `kubectl get pods --all-namespaces`  
```
NAMESPACE     NAME                                        READY     STATUS             RESTARTS   AGE
kube-system   etcd-kubespawndefault0                      1/1       Running            0          27m
kube-system   kube-apiserver-kubespawndefault0            1/1       Running            0          26m
kube-system   kube-controller-manager-kubespawndefault0   1/1       Running            0          27m
kube-system   kube-dns-545bc4bfd4-l944j                   0/3       Pending            0          27m
kube-system   kube-proxy-5bblr                            0/1       CrashLoopBackOff   9          27m
kube-system   kube-proxy-nxcd5                            0/1       CrashLoopBackOff   9          27m
kube-system   kube-scheduler-kubespawndefault0            1/1       Running            0          27m
kube-system   weave-net-7r68q                             1/2       CrashLoopBackOff   8          27m
kube-system   weave-net-d4j2p                             2/2       Running            9          27m
```
Not sure what I could be missing.
I ran the following commands after stopping my VM and starting it again.
```
sudo setenforce 0
cd $GOPATH/src/github.com/kinvolk/kube-spawn && sudo -E ./kube-spawn destroy
cd ~ && ./build.sh
```

dongsupark:
Hi Kosy, thanks for filing the issue.

I've also seen this case of kube-proxy (as well as kube-dns, weave-net) getting into crashing loops. It does not happen always. To figure out why it happens, it might be worth running `kubectl describe node kubespawndefault0`, which shows the current status of the node.

As I'm on holiday until the next week, I'm not sure I can look into kube-spawn right now.

dongsupark:
Today I had a chance to get back to this issue, and found out a bug when parsing kernel parameters. That's my fault. Sorry!
Please see https://github.com/kinvolk/kube-spawn/pull/262. Hopefully it will fix the issue.

[...]

PR #262, opened 2018-01-25
pkg/bootstrap: fix a bug when parsing kernel parameters

dongsupark:
`IsConntrackHashsizeCorrect()` is expected to return true, when the value of `/sys/module/nf_conntrack/hashsize` * 4 is less than the value of `/proc/sys/net/nf_conntrack_max`. However so far the string parsing logic has had a bug that the value obtained from the procfs or sysfs is passed without trimming whilespaces. As input string is usually given
such as `16384\n`, with a trailing newline character, `strconv.Atoi()` fails because of the newline. Then both `hs` and `ctmax` become 0, and `IsConntrackHashsizeCorrect()` returns always true. Obviously there will be no chance of adjusting the hashsize correctly as expected.

So we need to apply `strings.Trimspace()` for each input string. Then the newline character will be gone, and `IsConntrackHashsizeCorrect()` will work as expected.

Fixes https://github.com/kinvolk/kube-spawn/issues/255

### DIFF
--- a/pkg/bootstrap/node.go
+++ b/pkg/bootstrap/node.go
@@ -449,19 +449,29 @@ func isConntrackLoaded() bool {
 }
 
 func isConntrackHashsizeCorrect() bool {
-	hsStr, err := ioutil.ReadFile(ctHashsizeModparam)
+	hsByte, err := ioutil.ReadFile(ctHashsizeModparam)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


dongsupark:
As this is a critical bugfix, I'll merge it.
This PR is also independent of the refactoring PR https://github.com/kinvolk/kube-spawn/pull/265 , so there won't be merge conflicts.

PR #263, opened 2018-01-29
Dockerfile: build kube-spawn based on golang image instead of fedora

dongsupark:
For better build speed, we should use golang docker image instead of fedora, because we don't actually need the whole distro just for building go binaries.

Comparison of build duration, including fetching remote docker images:

* previously with fedora 27:   143.964 sec
* currently with golang latest: 74.608 sec

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,14 +1,7 @@
-FROM fedora:27
+FROM golang:latest
 
 ENV GOPATH /go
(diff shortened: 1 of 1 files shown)


dongsupark:
This PR is also independent of the refactoring PR https://github.com/kinvolk/kube-spawn/pull/265. So I'll merge it.

PR #265, opened 2018-02-01
Refactor kube-spawn

schu:
Not ready for review or merge yet. Still contains quite some *work in progress* code.

To test this separately from an existing kube-spawn setup, you can create a kube-spawn config file (yes, that's possible now) as follows:

```
sudo bash -c "mkdir -p /etc/kube-spawn; echo 'dir: /var/lib/kube-spawn-new' >/etc/kube-spawn/config.yaml"
```

Todos:

* [ ] Fix stop and destroy
* [ ] Fix multiprint


schu:
Note: I have removed support for Kubernetes 1.7 which is end of life in March IIRC.

schu:
A few example test examples:

To test with a custom Kubernetes build, build Kubernetes, push the hyperkube image to a registry of your choice (kube-spawn doesn't do that anymore) and give its location on the CLI, e.g.

```
sudo ./kube-spawn create \
  --hyperkube-image 10.22.0.1:5000/schu/hyperkube-amd64:schu-test \
  --kubernetes-source-dir ~/code/go/src/k8s.io/kubernetes \
  -c test-hyperkube

sudo ./kube-spawn start -n2 -c test-hyperkube
```

Note that `--kubernetes-source-dir` and you can mix kubeadm/kubelet from upstream with a custom hyperkube image (for example when you only want to test changes to the API).

To test with rktlet

```
sudo ./kube-spawn create \
  --container-runtime rkt \
  --rktlet-binary-path ~/code/go/src/github.com/kubernetes-incubator/rktlet/bin/rktlet

sudo ./kube-spawn start
```

All CLI flags can also be defined via environment variables or in the config file (and `--config` can be used to load a config file from a different location that `/etc/kube-spawn/config.yaml`), i.e. it's now easier to set this to preferred default values permanently.

dongsupark:
I'm still reading the code, but I want to say about 2 things.

First, the kubeadm options `--ignore-preflight-errors=all` and `--discovery-token-unsafe-skip-ca-verification` work only with kubeadm v1.9 or newer. It won't work at all with kubeadm v1.8. It's very likely that we should keep supporting k8s v1.8 for a while, so we cannot simply go with the new options.

Next, the machine name should not include dash (`-`). Please see https://github.com/kinvolk/kube-spawn/pull/144.

There are other things I want to point out, but I'd like to say about them altogether when this PR is ready for review. I like some parts, don't like some others.

I've just tested it with k8s v1.9.2. Basic things seem to work. It's not exactly bug-free, but let's fix the bugs together. 👍

schu:
> Next, the machine name should not include dash (-). Please see #144.

I would like to have "pretty names" and fix this. @alban what would I have to change to reproduce the peculiarities with dashes?

schu:
> First, the kubeadm options --ignore-preflight-errors=all and --discovery-token-unsafe-skip-ca-verification work only with kubeadm v1.9 or newer. It won't work at all with kubeadm v1.8. It's very likely that we should keep supporting k8s v1.8 for a while, so we cannot simply go with the new options.

I have re-added the `<1.9` option meanwhile.

dongsupark:
I've tested this PR today again. I think it's great to have possible kubeadm options back.
Though cluster names still have dashes.

It works mostly fine, but there are still bugs. One of the bugs is like that.

When destroying a cluster, the machines are not stopped at all, and its images stay as well. As a result, user needs to clean up everything by hand. Otherwise in the next phase, it's not possible to create the same cluster. This issue doesn't seem to always happen.

Anyway it's a drastic change, but I would like to merge this PR as soon as possible, because I have another patch to add on top of this one.

dongsupark:
For remaining issues regarding this PR, please create another separate issue.

PR #269, opened 2018-03-22
cmd, pkg/bootstrap: support Flatcar Linux as base image

dongsupark:
As the base image, we support [Flatcar Linux](https://flatcar-linux.org) images instead of CoreOS Container Linux images.

We replace hard-coded values "coreos" with a proper image name. Let's just remove the `--machinectl-image` option to make it simple.

### DIFF
--- a/pkg/bootstrap/node.go
+++ b/pkg/bootstrap/node.go
@@ -34,16 +34,21 @@ import (
 )
 
 const (
-	containerNameTemplate string = "kubespawn%d"
-	ctHashsizeModparam    string = "/sys/module/nf_conntrack/parameters/hashsize"
-	ctHashsizeValue       string = "131072"
-	ctMaxSysctl           string = "/proc/sys/net/nf_conntrack_max"
-	machinesDir           string = "/var/lib/machines"
-	machinesImage         string = "/var/lib/machines.raw"
-	coreosStableVersion   string = "1478.0.0"
+	containerNameTemplate  string = "kubespawn%d"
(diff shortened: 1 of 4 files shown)


pkg/bootstrap/node.go
```diff
@@ -639,15 +643,15 @@ func checkCoreosVersion() error {
...
+		if err := checkBaseImageVersionField(values); err != nil {
 			return err
 		}
 	}
 
 	return nil
 }
 
-func pullRawCoreosImage() error {
+func pullRawBaseImageImage() error {
```
schu: `s/pullRawBaseImageImage/pullBaseImage/`?

pkg/bootstrap/node.go
```diff
@@ -680,24 +684,24 @@ func pullRawCoreosImage() error {
...
+		log.Fatalf("You will need to remove the image by 'sudo machinectl remove %s' then the next run of kube-spawn will download version %s of coreos image automatically.", BaseImageName, baseImageStableVersion)
 	}
 }
 
-func PrepareCoreosImage() error {
-	// If no coreos image exists, just download it
-	if !machinectl.ImageExists("coreos") {
-		log.Printf("pulling coreos image...")
-		if err := pullRawCoreosImage(); err != nil {
+func PrepareBaseImageImage() error {
```
schu: `s/PrepareBaseImageImage/PrepareBaseImage/`?

dongsupark:
Updated the function names.
Thanks for review & test!

PR #271, opened 2018-03-23
Another fix for overlay check and for flags parsing

krnowak:
Commit messages say all.

### DIFF
--- a/pkg/bootstrap/util.go
+++ b/pkg/bootstrap/util.go
@@ -41,6 +41,10 @@ func PathSupportsOverlay(path string) error {
 		return fmt.Errorf("overlayfs is not available")
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/bootstrap/util.go
```diff
@@ -41,6 +41,10 @@ func PathSupportsOverlay(path string) error {
 		return fmt.Errorf("overlayfs is not available")
 	}
 
+	if err := os.MkdirAll(path, 0644); err != nil {
```
dongsupark: Do you probably mean `0755`?

krnowak: Ugh, yes. Will fix it.

krnowak:
PR updated.

dongsupark:
Looks good. Thanks!

PR #273, opened 2018-04-06
Disable kubelet eviction / allocatable enforcement

schu:
Also, fix a typo.


pkg/cluster/clusterfiles.go
```diff
@@ -87,14 +87,19 @@ systemctl start --no-block kubelet.service
 
 // --cgroups-per-qos should be set to false, so that we can avoid issues with
 // different formats of cgroup paths between k8s and systemd.
-// --enforce-node-allocatable= is also necessary.
+
+// --enforce-node-allocatable= and --eviction-hard= are set to make the
+// kubelot *not* evicting pods from the system, even when memory or disk
```
iaguis: kubelot

schu: Thanks, fixed.

PR #277, opened 2018-04-13
[WIP] Move `PathSupportsOverlay()` to `cluster.Start()`

schu:
Another leftover which should go into the cluster package.

Also, remove an old TODO, fix an error check plus some minor cleanup.


invidian:
@schu can we remove the branch here and in #257?

Issue #278, opened 2018-06-29
does not start with Kubernetes 1.11


dongsupark:
kube-spawn does not start when a cluster is created with Kubernetes v1.11.0, `--kubernetes-version=v1.11.0`.
`kube-spawn start` hangs after the following step:

```
Bootstrapping kube-spawn-default-master-ldgrrd ...
```

On the other hand, it works fine with k8s v1.10.
I suppose, it has something to do with changes in kubeadm. I haven't had a deeper look into it yet.

PR #279, opened 2018-06-29
pkg/cluster: fix bug when starting cluster with kubeadm >= 1.11

dongsupark:
When starting a kube-spawn cluster with kubeadm 1.11 or newer, `kube-spawn start` hangs forever during the bootstrap phase. It's because kubeadm 1.11 or newer asks users to type Y or n, while our bootstrap script runs only `kubeadm reset`. It works fine when running with
`kubeadm reset --force`, which breaks kubeadm 1.10 or older, because the lower versions don't support `--force`.

Fix the bug by dealing with kubeadm reset options differently when the kubeadm version is 1.11 or newer.

Fixes https://github.com/kinvolk/kube-spawn/issues/278

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -31,6 +31,7 @@ type ClusterSettings struct {
 	CNIPluginDir          string
 	ContainerRuntime      string
 	HyperkubeImage        string
+	KubeadmResetOptions   string
 	KubernetesSourceDir   string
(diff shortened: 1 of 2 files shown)


dongsupark:
Fixed typos in the commit message.
Tested with k8s 1.9, 1.10, and 1.11. It works fine.
Will merge.

PR #284, opened 2018-07-09
pkg/bootstrap: fix a race between machinectl image-status and umount

dongsupark:
There has been an issue of `umount /var/lib/machines` failing with -EBUSY. It happens only when a command `machinectl image-status` runs once at least. So we need to make a retry loop to attempt to umount, up to 3 seconds, to make it work correctly.


dongsupark:
Recently I haven't seen this issue.
Closing. Will revisit this issue in the future when it happens again.

Issue #280, opened 2018-06-29
Upgrade to v1alpha2 for kubeadm 1.12


dongsupark:
Starting from Kubernetes 1.11, the kubeadm configuration file version has been upgraded from `v1alpha1` to `v1alpha2`. `v1alpha1` read support exists in kubeadm v1.11, but will be [removed in v1.12](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#new-deprecations). 

So we should start to upgrade to `v1alpha2`, before the next Kubernetes release 1.12.

PR #287, opened 2018-07-19
pkg/cluster: change API to v1alpha2 for kubeadm >= 1.11

dongsupark:
Since k8s v1.11, API version of kubeadm has changed to `v1alpha2`, and `v1alpha1` is deprecated. So it's recommended to use `v1alpha2` for kubeadm >= 1.11, and `v1alpha1` will be completely removed in the upcoming release 1.12.

We need to introduce a new option for the API version, to conditionally deal with it.

Fixes https://github.com/kinvolk/kube-spawn/issues/280


dongsupark:
As k8s v1.12 is going to be released in a couple of weeks, we need to merge this PR.

PR #288, opened 2018-07-21
Add gofmt check

donbowman:
This PR fixes a few lint/fmt errors.
It then adds lint & fmt to the pipeline to prevent checking in code w/ fmt errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,13 @@ install: true
 go:
   - 1.9.x
 
+before_script:
+  - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b $GOPATH/bin v1.9.1
+
 script:
+  - echo run lint
+  - golangci-lint run --disable-all -E errcheck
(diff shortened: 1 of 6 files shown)


donbowman:
consider also running golangci-lint run $(go list ./... |sed -e 's?^_??' | grep -v /vendor/)

dongsupark:
I think, this PR fixes real bugs as well as missing error checks. So I'm totally fine with code itself. Thanks!

About updating travis config, I'm not 100% sure.

First of all, I'm fine with adding `gofmt` checks in the travis config.

On the other hand, as for adding `golint` checks in the travis config, I'm not sure if that's a good idea. Clearly running `golint` for a code base mostly helps us fix issues. Mostly it reports valid coding style fixes. But sometimes golint is too strict. For example `Api` vs `API`. So from my understanding, it's not always possible to fix every case reported by golint.

donbowman:
this is why i used errcheck mode of golint. its not style check.

dongsupark:
Aha, `golangci-lint` allows for running only a specific linter `errcheck`. That makes sense.
Then it looks good to me.

Issue #286, opened 2018-07-19
Use kubelet.conf instead of command-line options for kubelet


dongsupark:
Earlier in this week I had a look into the kubeadm API issue https://github.com/kinvolk/kube-spawn/issues/280.
Then I realized that there are a lot more upcoming deprecation issues, not in kubeadm but in kubelet.

Actually since K8s v1.10 the following kubelet options have been already deprecated, so it's recommended to use corresponding options in `kubelet.conf`. It's still working because it's still in the middle of the transition phase.

* `--authentication-token-webhook` (should be `authenticationTokenWebhook` in kubelet.conf)
* `--cgroup-driver` (should be `cgroupDriver` in kubelet.conf)
* `--cgroups-per-qos` (should be `cgroupsPerQOS` in kubelet.conf)
* `--container-runtime` (should be `containerRuntime` in kubelet.conf)
* `--container-runtime-endpoint` (should be `CRISocket` in kubelet.conf)
* `--eviction-hard` (should be `evictionHard` in kubelet.conf)
* `--enforce-node-allocatable` (should be `enforceNodeAllocatable` in kubelet.conf)
* `--fail-swap-on` (should be `failSwapOn` in kubelet.conf)
* `--runtime-request-timeout` (should be `runtimeRequestTimeout` in kubelet.conf)

See also https://github.com/kubernetes/kubernetes/pull/65191.

Not so urgent at the moment, but I'll try to create a PR to address this issue.

dongsupark:
Actually that transition has not happened upstream. https://github.com/kubernetes/kubernetes/pull/65191
Closing.

PR #292, opened 2018-07-27
[WIP] pkg/cluster: use kubelet.conf instead of cmdline parameters

dongsupark:
WIP

Since K8s v1.11, kubelet options should be defined in `kubelet.conf` instead of cmdline parameters. Still cmdline parameters are not deprecated yet, so it's not an urgent issue for now. Though we need to test this PR for possible scenarios.

Fixes https://github.com/kinvolk/kube-spawn/issues/286


dongsupark:
Actually that transition has not happened upstream. https://github.com/kubernetes/kubernetes/pull/65191
Closing.

PR #294, opened 2018-07-28
Add support for CNI plugin canal

donbowman:
Similar to flannel + calico, this will allow using canal.


donbowman:
any word on this one?

dongsupark:
I'll try to review it in this week.
It's just that other bug fixes have higher priorities.

dongsupark:
I've just tested this PR, and it looks good.
It works with both k8s v1.10 and v1.11.
Code is also simple enough to understand.

But why do we need the second commit?
"Merge remote-tracking branch 'upstream/master' into add-canal-cni"

donbowman:
the 2nd is just the merge from master so that this would apply clean, it should have no local changes

PR #295, opened 2018-08-05
Use os.Executable() to get our exe name rather than hard code.

donbowman:
kube-spawn might not be on the path, or there might be multiple,
or we may be using sudo which changes the path.


pkg/nspawntool/run.go
```diff
@@ -97,7 +101,7 @@ func Run(baseImageName, lowerRootPath, upperRootPath, machineName, cniPluginDir
 	defer stdout.Close()
 
 	if err := c.Start(); err != nil {
-		return errors.Wrap(err, "error running cnispawn")
+		return errors.Wrapf(err, "error running %s cnispawn: %s %v", ex, args)
```
dongsupark: 3 parameters are needed, but only 2 are given.

donbowman:
This changes means that no matter where you invoke kube-spawn from, its
used in the 2nd stage.

So if you run ./kube-spawn, it uses that. if you use /somewhere/kube-spawn,
it uses that. I think its what people expect.

schu:
IIRC we added `utils.Command` because in the past when kube-spawn was multiple binaries we wanted binaries in the current directory to take precedence, for example for easier debugging. Since kube-spawn is a single executable today and `utils.Command` only used in one place, it doesn't seem to be needed any longer.

> This changes means that no matter where you invoke kube-spawn from, its
used in the 2nd stage.

> So if you run ./kube-spawn, it uses that. if you use /somewhere/kube-spawn,
it uses that. I think its what people expect.

So I'm fine with the change, but it makes `utils.Command` obsolete I think, as the following code will never be true (or only when `os.Executable` returns an error, which I think we should rather return than ignore silently):

https://github.com/kinvolk/kube-spawn/blob/395c65f00d80d6d359197d81eee84d0a308d5c07/pkg/utils/command.go#L45

i.e. `pkg/utils/command.go` can be removed as far as I see.

donbowman:
are you asking me to remove that as part of the PR?

dongsupark:
The new commit looks good as well.
Tested again. Thanks! :+1:

PR #297, opened 2018-08-10
Document how to use kube-spawn on openSUSE Kubic

thkukuk:
kube-spawn works fine on openSUSE Kubic, document how to install it.


sysrich:
I can confirm the instructions work, but `kubectl` should be replaced with `kubernetes-client`

(kubectl is a package lurking around openSUSE:Tumbleweed to facilitate it's inclusion in SUSE's PackageHub; kubernetes-client is the package users should use in Kubic)

PR #300, opened 2018-08-16
Correct resize /var/lib/machines logic

donbowman:
This regression broke my ability to run (since /var/lib/machines is mounted on a real partition, not loopback).

Commit message has the bulk of the rationale.

Also correct missing error handling.

### DIFF
--- a/pkg/bootstrap/node.go
+++ b/pkg/bootstrap/node.go
@@ -93,13 +93,59 @@ func GetPoolSize(baseImageName string, nodes int) (int64, error) {
 	return poolSize, nil
 }
 
+func setPoolLimit(poolSize int64) error {
+	var cmdPath string
+	var err error
+
+	if cmdPath, err = exec.LookPath("machinectl"); err != nil {
+		return fmt.Errorf("machinectl not installed: %s", err)
+	}
+
+	args := []string{
+		cmdPath,
+		"set-limit",
+		strconv.FormatInt(poolSize, 10),
+	}
+
+	cmd := exec.Cmd{
+		Path:   cmdPath,
+		Args:   args,
+		Env:    os.Environ(),
+		Stdout: os.Stdout,
+		Stderr: os.Stderr,
+	}
+
+	if err := cmd.Run(); err != nil {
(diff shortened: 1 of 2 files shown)


dongsupark:
Thanks for this PR.
I think most points addressed by this PR are valid.
Though I think it is a little difficult to review.
It has commits from another PR for canal support, which should be independent of this PR.
The last commit itself, which includes most valid bug fixes, also tries to fix quite a lot different topics at the same time.
It would be great if it were split out into multiple PRs or multiple commits.

donbowman:
Ooops.
Let me fix the canal mistake.

I'm not sure how to split the last commit tho?

donbowman:
I have split the PR into 2.

This one (with the 'canal' change backed out) is just the fix for assuming /var/lib/machines.raw is existing and loopback.

The other will deal with the missing error handling and assuming host paths / selinux.

pkg/bootstrap/node.go
```diff
@@ -93,13 +93,59 @@ func GetPoolSize(baseImageName string, nodes int) (int64, error) {
...
+	if err := syscall.Statfs(machinesDir, &stat); err == nil {
+		if stat.Bsize >= poolSize {
+			return nil
+		}
+	}
+
+	// First call `machinectl set-limit size`. If this succeeds,
+	// we are done
+	if err := setPoolLimit(poolSize); err == nil {
+		return nil
```
dongsupark: Usually Go's error handling is done the other way around, like:

```
if err != nil {
        return err
}
```

I feel awkward whenever I see the opposite way of error handling.
I know it's sometimes hard to handle errors in a correct way, and I'm not sure if we should always force everyone to write in that way.
So I think that the code is fine for now.
But in the future, it would be really great, if you could follow the idiomatic coding style. :-)

donbowman: in this case its not checking for error, its checking for success. E.g. if setPoolLimit() succeeds, we are done. Only if it fails do we need to try the backup 'unmount/resize/remount'.

e.g. i would not want to return err @ this time.

PR #301, opened 2018-08-20
Add error handling to utility functions, remove assumptions about host

donbowman:
Most error checking was missing from pkg/bootstrap/node.go.
Added to all utility functions.

In addition, pkg/bootstrap/node.go made assumptions about
host-path for binaries (machinectl/qemu-img/mount etc). Remove this.
If the binaries are not on the PATH, fail. This is the reason
for Unix PATH, to indicate where to find executables, hard-coding
fallbacks can create a security issue or astonishment.

The code had a hard-coded assumption that the machine had
selinux available (e.g. it assumed selinux is enforcing/not-enforcing)
where there is a 3rd-option (not available). If selinux is not
available (e.g. Ubuntu-derived hosts), then return false
from isSELinuxEnforcing().

### DIFF
--- a/pkg/bootstrap/node.go
+++ b/pkg/bootstrap/node.go
@@ -136,8 +136,7 @@ func runImageResize(poolSize int64) error {
 	var err error
 
 	if cmdPath, err = exec.LookPath("qemu-img"); err != nil {
-		// fall back to an ordinary abspath to qemu-img
-		cmdPath = "/usr/bin/qemu-img"
+		return fmt.Errorf("qemu-img not installed: %s", err)
 	}
 
 	args := []string{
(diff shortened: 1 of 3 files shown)


pkg/bootstrap/node.go
```diff
@@ -329,60 +331,55 @@ func runModprobe(moduleName string) error {
...
-	if isOverlayfsAvailable() {
-		return
-	}
+func ensureOverlayfs() error {
+	if !isOverlayfsAvailable() {
 
-	log.Println("Warning: overlayfs not found, docker would not run.")
-	log.Println("loading overlay module... ")
+		// This is an incorrect assumption. It depends on the docker
+		// storage driver
```
dongsupark: In this function, code looks effectively similar as before.
But I'm not sure I understand this comment.
Are you saying that we should check for docker storage driver in the future?
In that case, we should create another github issue, or add `TODO` to the comment.

donbowman: issue added
https://github.com/kinvolk/kube-spawn/issues/302

dongsupark: Thanks for adding the issue.

PR #305, opened 2018-08-30
Make master/worker be numbered rather than random-string for name.

donbowman:
Rationale: using NodePort, its simpler to remember `kube-spawn-cluster-worker-1` than to do a machinectl and find the machine name which was used. The former is invariant across invocations.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -391,9 +391,9 @@ func (c *Cluster) Start(numberNodes int, cniPluginDir string, cniPlugin string) 
 
 			var machineNameSuffix string
 			if nodeNumber == 0 {
-				machineNameSuffix = fmt.Sprintf("master-%s", randString(6))
+				machineNameSuffix = fmt.Sprintf("master-%d", nodeNumber)
 			} else {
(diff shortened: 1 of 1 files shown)


dongsupark:
Hmm, I'm not sure about this.
I agree that it would be simpler to remember with human-readable strings. That was our original behavior, until we recently took the current approach. If we go back to the original behavior again, then we would lose uniqueness in the names.
Let me think.

donbowman:
the names are still unique.

kube-spawn-<CLUSTER>-<master|worker>-#

and # is incrementing.

it just allows me to always refer to a worker when using NodePort to come
into it, rather than run machinectl to find the IP.

donbowman:
any comment on this?

dongsupark:
@donbowman Sorry. This PR slipped through my fingers.

How about making a command-line option for the different naming scheme, instead of changing the default one?
For example, `--sequential-node-names`.
If that option is given, we use the new human-readable naming scheme.
If not, we stay with the current randomized names.
Would it make sense?

donbowman:
I guess i don't understand what the random one achieves for anyone? Its no more unique.

PR #306, opened 2018-08-30
pkg/bootstrap: enlarge pool size also before pulling base image

dongsupark:
If the storage pool is not initialized at all, the storage pool could be only 500MB by default, which is way smaller than the default image size of Flatcar Linux, 1.6G.

So we need to resize the pool to around 2G, even before pulling the base image.


dongsupark:
It's really needed on every test. Merging.

PR #314, opened 2018-09-05
Enable ipv4/ipv6 dual-stack in Calico.

donbowman:
Calico is capable of enabling ipv4+ipv6 dual stack usage, but only if it has an ipv6 range in ipam. Without the ipv6 range, even lo is not bound to a v6 address, preventing code which explicitly wants to use v6 even locally (e.g. unit tests) from operating.

Add in an explicit v6 range (from the the unique local address range of fc00::/7). Do this by mirroring the config file from

 https://docs.projectcalico.org/v3.1/getting-started/kubernetes/installation /hosted/kubernetes-datastore/calico-networking/1.7/calico.yaml

into the system, with the change. The ipv4 section of the ipam is configured to match Calico's felix expectation.

### DIFF
--- a/pkg/cluster/clusterfiles.go
+++ b/pkg/cluster/clusterfiles.go
@@ -143,3 +143,494 @@ for arg in "${@}"; do
 done
 exec docker-runc "${args[@]}"
 `
+
+// This matches "https://docs.projectcalico.org/v3.1/getting-started/kubernetes/installation
+//               /hosted/kubernetes-datastore/calico-networking/1.7/calico.yaml"
+// except that the ipam is overriden to use 2 ranges, 1 as expected
+// in Calico felix, and one for ipv6. This has the affect of enabling dual-stack
+// so that code in a Pod can bind to ::1 for lo.
+const CalicoNet = `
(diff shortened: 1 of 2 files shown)


pkg/cluster/clusterfiles.go
```diff
@@ -143,3 +143,494 @@ for arg in "${@}"; do
...
 exec docker-runc "${args[@]}"
 `
+
+// This matches "https://docs.projectcalico.org/v3.1/getting-started/kubernetes/installation
+//               /hosted/kubernetes-datastore/calico-networking/1.7/calico.yaml"
+// except that the ipam is overriden to use 2 ranges, 1 as expected
+// in Calico felix, and one for ipv6. This has the affect of enabling dual-stack
+// so that code in a Pod can bind to ::1 for lo.
+const CalicoNet = `
+# Calico Version v3.1.3
```
dongsupark: Aha, the yaml file was taken from projectcalico, and modified to supporting IPv6.
Hmm, I'm not sure I like the way of including the whole file in the source code.
But I don't have a better idea.

PR #316, opened 2018-09-12
pkg/bootstrap: verify hash of each k8s binary file

dongsupark:
Downloaded k8s binary files, e.g. `kubectl`, sometimes can be corrupted, due to many reasons, like intermittent network errors. The thing is, in this case kube-spawn simply continues to start clusters with the broken files, and actually tries to run those. It obviously fails, but the failures are not directly visible to users, so users need to get into one of the containers to debug the issues. For most users it's quite hard.

To fix that, we should verify if each downloaded Kubernetes binary file is valid, with help of the given sha1 hash file, i.e. `filename+".sha1"`. (Unfortunately on dl.k8s.io, there's no other hash file available, no sha256 at all) If the hash is not correct, simply error out to inform users that something is wrong with the files.

For example, it does the same job as below:
```
$ curl -sSLO https://dl.k8s.io/v1.11.0/bin/linux/amd64/kubectl
$ curl -sSLO https://dl.k8s.io/v1.11.0/bin/linux/amd64/kubectl.sha1
$ [ "$(sha1sum kubectl | cut -d\  -f1)" = "$(cat kubectl.sha1)" ] && echo "pass" || echo "fail"
pass
```

### DIFF
--- a/pkg/bootstrap/download.go
+++ b/pkg/bootstrap/download.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/kinvolk/kube-spawn/pkg/utils"
 	"github.com/kinvolk/kube-spawn/pkg/utils/fs"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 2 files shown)


blixtra:
@dongsupark Should we not file an issue upstream about that?

blixtra:
On the missing sha1 hash.

dongsupark:
Filed an upstream issue: https://github.com/kubernetes/kubernetes/issues/68555

PR #320, opened 2018-10-04
pkg/cluster: add kind field in the kubeadm config option

dongsupark:
Starting from kubeadm v1.12.0, both `kind` and `apiVersion` are necessary in the kubeadm config file when initializing the cluster. If `kind` is missing, `kubeadm init` simply fails like this:

```
master-ajk8ix invalid configuration: kind and apiVersion is mandatory information that needs to be specified in all YAML documents
Failed to start cluster ...
```


dongsupark:
Last week I tested this PR with k8s v1.9, v1.10, v1.11, and v1.12.
Will merge as it's necessary for running with v1.12.

Issue #326, opened 2018-10-27
privileged operation doesn't work


donbowman:
If I create a privileged pod running dind (as eg we might do if using gitlab-runner):

```
kubectl apply -f - << EOF
apiVersion: v1                                                                                                                                               
kind: Pod
metadata:
  name: dind
  labels:
    role: dind
spec:
  containers:
    - name: dind
      image: docker:dind
      securityContext:
        privileged: true
EOF
```

And then I kubectl exec into it and run:

```
# docker pull alpine
# docker run --rm -it alpine sh               
```
It ends up not working:
```                                                                                                             
docker: Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "process_linux.go:402: container init caused \"could not create session key: operation not permitted\"": unknown.
```

In the dind container we see this log:

```
time="2018-10-27T21:07:10Z" level=info msg="shim docker-containerd-shim started" address="/containerd-shim/moby/7428063ad54a9fafb5bf377d36f2f2c977472b67ae0001854c6e71c533a22f27/shim.sock" debug=false pid=305 
time="2018-10-27T21:07:10Z" level=info msg="shim reaped" id=7428063ad54a9fafb5bf377d36f2f2c977472b67ae0001854c6e71c533a22f27                                 
time="2018-10-27T21:07:10.253654428Z" level=error msg="stream copy error: reading from a closed fifo"                                                        
time="2018-10-27T21:07:10.498102350Z" level=error msg="7428063ad54a9fafb5bf377d36f2f2c977472b67ae0001854c6e71c533a22f27 cleanup: failed to delete container from containerd: no such container"
time="2018-10-27T21:07:10.512689130Z" level=error msg="Handler for POST /v1.38/containers/7428063ad54a9fafb5bf377d36f2f2c977472b67ae0001854c6e71c533a22f27/start returned error: OCI runtime create failed: container_linux.go:348: starting container process caused \"process_linux.go:402: container init caused \\\"could not create session key: operation not permitted\\\"\": unknown"
```

donbowman:
This is because: `# echo 123 | keyctl padd user foobar @u`
fails in the systemd-nspawn'd containers with                          
`add_key: Operation not permitted`

dongsupark:
https://github.com/kinvolk/kube-spawn/pull/328 was merged, so I suppose your issue would be gone.
If there's still the same issue, feel free to reopen it.

PR #329, opened 2018-11-25
Enable keyring and bind-mount /sys/kernel/security for privileged operation

donbowman:
This resolves part of the issue referenced in
[#326](https://github.com/kinvolk/kube-spawn/issues/326) needed to run
privileged pods. However, I don't believe it is sufficient to fully
support a Pod which wants to run e.g. Docker-In-Docker.

Signed-off-by: Don Bowman <db@donbowman.ca>


dongsupark:
Thanks a lot!
I've just tested it, and it works fine. :+1: 

Nit: we don't necessarily require `Signed-off-by` in the commit message. Anyway I think it's fine either way.

PR #332, opened 2018-12-05
Fix issues for running with k8s v1.13

dongsupark:
Let's bump the default Kubernetes version to v1.12.3, the most secure up-to-date version. This is mostly because of the critical security issue https://github.com/kubernetes/kubernetes/issues/71411.

Apply CNI plugins outside of nspawn containers. Starting from k8s v1.13, the insecure port 8080 of kube-apiserver is not available any more, as it has been deprecated. So we cannot run `kubectl apply` inside of nspawn containers.

Kubeadm v1.13 or newer should run with a config with `apiVersion: v1beta1`. In case of apiVersion `v1alpha*`, the `InitConfiguration` part can contain `apiServerExtraArgs`, `authorizationMode`, and `controllerManagerExtraArgs`. However, in case of apiVersion  `v1beta*`, these option should be removed or moved into each corresponding part. Otherwise kubeadm would complain with json unmarshal errors.

### DIFF
--- a/cmd/kube-spawn/create.go
+++ b/cmd/kube-spawn/create.go
@@ -47,7 +47,7 @@ func init() {
 	kubespawnCmd.AddCommand(createCmd)
 
 	createCmd.Flags().String("container-runtime", "docker", "Runtime to use for the cluster (can be docker or rkt)")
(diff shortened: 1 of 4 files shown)


dongsupark:
Will merge, as it's necessary for running kube-spawn with k8s 1.13.

PR #343, opened 2019-05-21
Dockerfile/Makefile: set GOCACHE to a shared volume on the host

dongsupark:
Now that Go 1.12 or newer requires `GOCACHE`, a dockerized build fails with `permission denied` when initializing gocache.

To avoid such failures we should make a shared volume for the go cache to make it
available inside the container.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ DOCKER_TAG ?= latest
 PREFIX ?= /usr
 BINDIR ?= ${PREFIX}/bin
 UID=$(shell id -u)
+GOCACHEDIR=$(shell which go >/dev/null 2>&1 && go env GOCACHE || echo "$(HOME)/.cache/go-build")
 
 .PHONY: all clean dep install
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -12,8 +12,11 @@ VERSION=$(shell git describe --tags --always --dirty)
 ifeq ($(DOCKERIZED),y)
 all:
 	docker build -t kube-spawn-build:$(DOCKER_TAG) -f Dockerfile.build .
+	mkdir -p /tmp/.cache
+	chown -R $(UID):$(UID) /tmp/.cache
 	docker run --rm -ti \
 		-v `pwd`:/go/src/github.com/kinvolk/kube-spawn:Z \
+		-v /tmp/.cache:/tmp/.cache:Z \
```
invidian: How about standard `~/.cache/go-build/`? Or even output of `go env GOCACHE`?

dongsupark: Good point. I've changed the host cache directory to `GOCACHE`.

Makefile
```diff
@@ -4,6 +4,7 @@ DOCKER_TAG ?= latest
 PREFIX ?= /usr
 BINDIR ?= ${PREFIX}/bin
 UID=$(shell id -u)
+GOCACHEDIR=$(shell go env GOCACHE)
```
invidian: Actually, this requires `go` binary on the host. However, I think it's a sane requirement.
Alternatively, we can do following:
```
GOCACHEDIR=$(shell which go >/dev/null 2>&1 && go env GOCACHE || echo "$(HOME)/.cache/go-build"
```

invidian (approved):
Tested and it's working fine. I found one possible improvement, but I don't insist on it.

Makefile
dongsupark: Ok, done.

Issue #339, opened 2019-02-19
cluster fails to start when running with systemd >= v240


dongsupark:
When trying to test kube-spawn with systemd v240 or newer, it does not start the whole cluster.

Only one of the nodes becomes active, and only the node gets registered to systemd-machined. Actually all of the nodes should be active. Which node becomes active seems to be pretty random. Sometimes the master is active, sometimes one of the workers is active. In some cases, every node becomes inactive.

On the other hand, when running with systemd v239, everything works fine.

I have not managed to track down its reason. Probably git-bisect between v239 and v240 is needed.

invidian:
https://github.com/systemd/systemd/blob/401faa3533280b05fee972e0c64885caf4b31e4c/units/var-lib-machines.mount#L10 this might be the reason why it's not spawning.

When I try to spawn it on Ubuntu 19.04 with systemd version 240.

I get this error:
```
    ubuntu: Download of https://alpha.release.flatcar-linux.net/amd64-usr/current/flatcar_developer_container.bin.bz2 complete.
    ubuntu: Failed to set file attributes on /var/lib/machines/.#flatcar.raw43b38c6ebee86770: Operation not supported
    ubuntu: Created new local image 'flatcar'.
    ubuntu: Operation completed successfully.
    ubuntu: Exiting.
    ubuntu: + sudo GOPATH=/home/vagrant/go ./kube-spawn create --cni-plugin-dir=/home/vagrant/go/bin
    ubuntu: Downloading kubectl
    ubuntu: Downloading 10-kubeadm.conf
    ubuntu: Downloading kubelet.service
    ubuntu: Downloading kubelet
    ubuntu: Downloading kubeadm
    ubuntu: downloading socat
    ubuntu: Copying files for cluster ...
    ubuntu: Generating configuration files from templates ...
    ubuntu: Cluster default created
    ubuntu: + sudo GOPATH=/home/vagrant/go ./kube-spawn start --cni-plugin-dir=/home/vagrant/go/bin --nodes=2
    ubuntu: Warning: kube-proxy could crash due to insufficient nf_conntrack hashsize.
    ubuntu: setting nf_conntrack hashsize to 131072... 
    ubuntu: making iptables FORWARD chain defaults to ACCEPT...
    ubuntu: Failed to start cluster: stat /var/lib/machines.raw: no such file or directory
```

I'm testing current workaround now:
```diff
diff --git a/pkg/bootstrap/node.go b/pkg/bootstrap/node.go
[...]

PR #346, opened 2019-05-22
misc: migrate from dep to go modules

### LINKED ISSUE #345: Migrate from dep to go modules
Go modules is now a standard way of managing dependencies, so we should switch to it to make building process easier.

invidian:
Closes #345

Note: this should probably be rebased once #343 is merged and then reviewed to make sure CI passes.


invidian:
Oops, it seems that Fedora 28, which is used by default ships go 1.10, which don't have modules support. I'll update defaults to use Fedora 30 and check if everything works as expected still.

dongsupark:
I've tested on my local system, Fedora 30. It fails with:

```
$ DOCKERIZED=n make
go mod download
go build -ldflags "-X main.version=v0.3.0-37-g24499024" \
        ./cmd/kube-spawn
# github.com/kinvolk/kube-spawn/pkg/cnispawn
pkg/cnispawn/netns.go:38:16: undefined: ns.NewNS
make: *** [Makefile:26: all] Error 2
```

I'm not completely sure, but with CNI plugins 0.7 or newer, we cannot call `NewNS` which does not exist, right?
No wonder, because kube-spawn has never been tested with newer CNI plugins like 0.7.

invidian:
@dongsupark see my commit message:
```
Note: github.com/containernetworking/plugins must be v0.7.0 as from
version v0.8.0 they removed code resposible for network namespace
creation.

See
containernetworking/plugins@a0eac8d#diff-d3d859d29b12afe98694dc54a7ced6d9
for details.
```

It also builds in Vagrant, so it must be your local environment issue. I suggest executing `GO111MODULE=on go clean -modcache`.

In any case, also Vagrantfile and scripts require changing to use go modules, so this PR is incomplete.

Makefile
```diff
@@ -5,8 +5,9 @@ PREFIX ?= /usr
 BINDIR ?= ${PREFIX}/bin
 UID=$(shell id -u)
 GOCACHEDIR=$(shell which go >/dev/null 2>&1 && go env GOCACHE || echo "$(HOME)/.cache/go-build")
+GO111MODULE=on
```
dongsupark: I think the issue is that the line above does not set the env variable `$GO111MODULE` expected.

So if I manually set `GO111MODULE=on` for the go build command, it works fine.
On the other hand, if I run it without `GO111MODULE=on`, it sometimes fails, because it tries to look for the cni plugins repo under my GOPATH. If my local cni plugins repo is based on 0.7, it works fine. If not, it fails.
That's not exactly what users would expect to happen.

invidian: Ah, that's my missing knowledge regarding `Makefile`. Should I add `export GO111MODULE=on` to `all` and `update-vendor` then?

dongsupark: Yes, you can do so.
Or you can simply add `GO111MODULE=on` only for go build commands.
If it works, I don't care much.

invidian:
I pushed fixed `Makefile`. However, there are still some changed pending, which needs to be pushed (updating Fedora release etc).

Makefile
```diff
@@ -15,21 +15,22 @@ all:
...
 		--user $(UID):$(UID) \
 		kube-spawn-build
 else
 all:
-	go build -o kube-spawn \
-		-ldflags "-X main.version=$(VERSION)" \
+	export GO111MODULE=on
+	go mod download
+	go build -ldflags "-X main.version=$(VERSION)" \
 		./cmd/kube-spawn
```
dongsupark: Sorry, I was not specific.
The current Makefile still does not set `GO111MODULE` correctly.

In Makefile, you need to specify env variables either in the same line you want to run, or in one of the multiple lines separated by `;`.

For example you can do either

```
        go mod download
        GO111MODULE=on go build
```

or

```
        export GO111MODULE=on; \
        go mod download; \
        go build
```
See also https://stackoverflow.com/questions/42993169/setting-environment-variable-in-makefile-and-reading-it-from-python-script-does

Makefile
```diff
@@ -15,21 +15,22 @@ all:
...
 endif
 
-update-vendor: | dep
-	dep ensure
-dep:
-	@which dep || go get -u github.com/golang/dep/cmd/dep
+update-vendor:
+	export GO111MODULE=on
+	go get -u
+	go mod tidy
```
dongsupark: Ditto

dongsupark (commented):
See inline comments:

Makefile
invidian: Yes, I have this one fixed already, just didn't push it yet. Testing takes time...

dongsupark (approved):
LGTM in general.

First of all, `go modules` works well as expected. Great!

Vagrant test scripts work as well, both for Fedora and Ubuntu.

Though I'm still seeing the issue of `kube-spawn start` hanging, so I'm not sure this PR would fix https://github.com/kinvolk/kube-spawn/issues/339. Anyway I'm fine with merging this PR, as that looks like a separate issue.

And in general I would like to see multiple separate PRs, instead of 16 commits in a single PR. Having an individual PR for each topic would make reviews a lot easier.

Minor issues inline: (Though I don't have strong opinions)

Vagrantfile
```diff
@@ -45,10 +45,10 @@ Vagrant.configure("2") do |config|
     end
   end
 
-  # Ubuntu 18.04 (Artful)
+  # Ubuntu 19.04 (Disco)
   config.vm.define "ubuntu", autostart: false do |ubuntu|
-    config.vm.box = "generic/ubuntu1804"
-    config.vm.provision "shell", inline: "apt-get update; DEBIAN_FRONTEND=noninteractive apt-get install -y btrfs-progs git golang iptables make policykit-1 qemu-utils rinetd selinux-utils systemd-container"
+    config.vm.box = "generic/ubuntu1904"
+    config.vm.provision "shell", inline: "apt-get update; DEBIAN_FRONTEND=noninteractive apt-get install -y btrfs-progs git golang-1.11 iptables make policykit-1 qemu-utils rinetd selinux-utils systemd-container"
```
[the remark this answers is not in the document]
invidian: I picked 1.11 as it's the same version shipped by Fedora 29 and Debian testing. I upgraded to Fedora 30 first, then rolled back to Fedora 29 to see if that solves cluster spawning issues, then updated to Fedora 30 again. We can bump this to 1.12.

vagrant-all.sh
```diff
@@ -8,7 +8,7 @@ export KUBESPAWN_REDIRECT_TRAFFIC="true"
 
 KUBESPAWN_PROVIDER=${KUBESPAWN_PROVIDER:-virtualbox}
 
-vagrant up $KUBESPAWN_DISTRO --provider=$KUBESPAWN_PROVIDER
+vagrant up $KUBESPAWN_DISTRO --provider=$KUBESPAWN_PROVIDER ${KUBESPAWN_VAGRANT_EXTRA_FLAGS}
```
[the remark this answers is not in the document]
invidian: We don't because we don't do `set -u`. However, it's a good idea to add both (declaration and -u), so I'll do it.

invidian:
Thanks for the review and feedback @dongsupark.

>And in general I would like to see multiple separate PRs, instead of 16 commits in a single PR. Having an individual PR for each topic would make reviews a lot easier.

I know amount of commits started piling up and I'm not the fan of it either. Unfortunately the more I was testing, the more possible improvements I was fining and implementing. I can still split it into separated PRs if you like.

>Though I'm still seeing the issue of kube-spawn start hanging, so I'm not sure this PR would fix #339. Anyway I'm fine with merging this PR, as that looks like a separate issue.

You are right about this. By the time of writing, I assumed this was the only issue with systemd >= 240, turns out there are some other issues with networking, which I didn't have time to investigate yet. I changed commit message to only reference to these issues.

I addressed your comments, so PR is ready for another round of review.

dongsupark:
Looks good. Thanks!

